PR #65, opened 2025-07-02
Add new guest-opensuse-tumbleweed image

LakshmiSaiHarika:
Added OpenSUSE tumbleweed  guest image with minimum package list that supports QEMU boot


LakshmiSaiHarika:
I will revisit this PR after I finish logging feature

PR #77, opened 2025-08-05
Refactor snphost module to capture snphost ok status(pass/failure) in the summary

LakshmiSaiHarika:
This PR updates the snphost OK service status to update JOB_RESULT json attribute of journalctl json log to "done" or "failed." 

Previously, the journalctl JSON attribute "JOB_RESULT" would always show "done" regardless of whether the snphost test passed or failed.
With this PR, snphost failure is captured in journalctl json log

PR #78, opened 2025-08-06
Refactor snpguest service type

LakshmiSaiHarika:
Converted all snpguest related service types to `oneshot` to perform one-time action and exit with the actual status (pass/fail)

This PR contains the commit of PR #77. 
Please review this PR after PR #77

### DIFF
--- a/modules/guest-measurement/mkosi.extra/usr/local/lib/systemd/system/calculate-measurement.service
+++ b/modules/guest-measurement/mkosi.extra/usr/local/lib/systemd/system/calculate-measurement.service
@@ -1,9 +1,8 @@
 [Unit]
 Description=Calculate guest measurement
 DefaultDependencies=no
-After=network-online.target snphost-ok.service
+After=network-online.target
 Wants=network-online.target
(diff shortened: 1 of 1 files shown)


modules/guest-measurement/mkosi.extra/usr/local/lib/systemd/system/calculate-measurement.service
```diff
@@ -1,9 +1,9 @@
 [Unit]
 Description=Calculate guest measurement
 DefaultDependencies=no
-After=network-online.target snphost.service
+After=network-online.target snphost-ok.service
```
DGonzalezVillal, line 4: In my work, I removed the dependency of snphost-ok for calculate measurement. I thought about it, and you can still calculate the measurement even if the snphost ok fails. 

They both need to be dependencies to launch the guest

LakshmiSaiHarika: Ok, I will remove this snphost-ok.service dependency then

DGonzalezVillal (approved):

PR #79, opened 2025-08-08
Read guest journal on host

LakshmiSaiHarika:
I added a new service to read the guest journal on the host to check for systemd-journal-remote service status on the host and  test for guest log upload process via systemd-journal-upload service running on guest.

Added the service dependency order as follows:
   read-guest-journal.service --> show-snp-guest-test.service --> show-snp-guest-attestatiion.service


LakshmiSaiHarika:
Screenshots for the failed read-guest-journal.service status and skipped show-snp-guest-test service on Ubuntu Plucky  are as follows:
image

PR #84, opened 2025-09-09
Feat: Add debian support into snpcert

LakshmiSaiHarika:
- Removed Debian Bookworm distribution, as it has kernel 6.1 version which doesn't support SNP feature.
- Added Debian Trixie distribution, which comes with the latest kernel version(6.12) that supports SNP feature


LakshmiSaiHarika:
Please review the Debian trixie certification logging results on the below link:
[Debian Logging Result](https://docs.google.com/document/d/1CfX_4CUKwgY6bIBSOvLgDz1S1zcY092OVGbP7voEj8U/edit?tab=t.qw0i20r8i44g
)

DGonzalezVillal (approved):

PR #86, opened 2025-09-11
Add centos support

LakshmiSaiHarika:
- Removed `ubuntu-oracular` as this release has reached End of Life, and no longer receives any security updates from ubuntu upstream.
- Added host/guest centos stream version 10 support for SNP certification process
- To fix the networking issue on centos, I added `systemd-networkd` build and downloaded the resolv.conf from systemd upstream, because the direct symlink to/etc/resolv.conf from centos systemd-resolved pkg fails during mkosi.build with the error `/etc/resolv.conf: Resource/device busy`.


modules/launch-snp-guest/mkosi.extra/usr/local/lib/scripts/launch-guest.sh
```diff
@@ -29,6 +29,7 @@ truncate -s 0 ${GUEST_ERROR_LOG}
 # Launch the SNP guest in background
 exec qemu-system-x86_64 \
   -enable-kvm \
+  -machine q35 \
```
DGonzalezVillal, line 32: Why did you have toa dd this, did you verify it didn't affect other distros?

DGonzalezVillal: you need it for centos?

LakshmiSaiHarika: Yes, I added it as qemu-machine-type fix for centos

PR #88, opened 2025-09-17
Adding Beacon support to SNPCERT

DGonzalezVillal:
This PR adds beacon support to SNPCERT:

- Fix the hanging issue we were seeing with beacon
- Introduces the beacon module
- Adds the necessary changes and dependencies so that our existing images can use beacon


modules/beacon/mkosi.extra/usr/local/lib/systemd/system/enable-beacon.service
```diff
@@ -0,0 +1,16 @@
...
+After=network-online.target avahi-daemon.service
+Wants=network-online.target
+Requires=avahi-daemon.service
+
+[Service]
+Type=simple
+ExecStart=beacon boot
+StandardOutput=journal
+StandardError=journal
+ExecStartPost=/usr/bin/touch /run/enable-beacon.ok
```
npmccallum, line 13: What is the purpose of /run/enable-beacon.ok?

modules/beacon/mkosi.extra/usr/local/lib/systemd/system/publish-certificate.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Publish SEV certificate using beacon
+DefaultDependencies=no
+After=sev-certificate-generator.service
+Wants=sev-certificate-generator.service
+Requires=enable-beacon.service
```
npmccallum, line 6: We probably want Requires/After for beacon-boot.service here. Also, we should probably add the avahi requirement too.

modules/beacon/mkosi.extra/usr/local/lib/systemd/system-preset/10-beacon.preset
```diff
@@ -0,0 +1,3 @@
+enable avahi-daemon.service
+enable enable-beacon.service
+enable publish-certificate.service
```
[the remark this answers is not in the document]
DGonzalezVillal: Yeah I think during testing I hadn't add the required so I was doing in manually here, but you are right with required it should be enough.

beacon/src/avahi/mod.rs
```diff
@@ -34,13 +36,10 @@ impl<'a> Avahi<'a> {
         Ok(browsing::Browsing(proxy.receive_item_new().await?))
     }
 
-    pub async fn resolve(
-        &self,
-        service: browsing::Service,
-    ) -> Result<resolving::Resolving, zbus::Error> {
-        let path = self
+    pub async fn resolve(&self, service: browsing::Service) -> Result<Resolved, zbus::Error> {
+        let (iface, proto, name, stype, domain, host, aprot, addr, port, txt, flags) = self
```
npmccallum, line 40: `let values = ...`

We only ever refer to this tuple as a whole because of the TryFrom impl below. Therefore, we only need a single variable here.

beacon/src/avahi/mod.rs
```diff
@@ -51,11 +50,19 @@ impl<'a> Avahi<'a> {
...
+            name,
+            stype,
+            domain,
+            host,
+            aprot,
+            addr.clone(),
+            port,
+            txt,
+            flags,
+        ))
```
npmccallum, line 65: `values.try_into().map_err(...)`

beacon/src/avahi/resolving.rs
```diff
@@ -15,40 +12,67 @@ pub struct Resolved {
...
+        String,       //name 2
+        String,       //service_type 3
+        String,       //domain 4
+        String,       //host 5
+        i32,          //approtocol 6
+        String,       //address 7
+        u16,          //port 8
+        Vec<Vec<u8>>, //txt 9
+        u32,          //flags 10
+    )> for Resolved
```
npmccallum, line 28: This is a long, complex type. A type alias might be better than bundling all these values in the impl. I wish this was a struct. But I can't blame zbus for that.

beacon/src/main.rs
```diff
@@ -135,36 +135,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+            Ok(resolved) => {
                 match resolved.address.ip() {
                     addr if addr.is_loopback() => continue,
                     IpAddr::V4(ipv4) if ipv4.is_link_local() => continue,
                     IpAddr::V6(ipv6) if ipv6.is_unicast_link_local() => continue,
                     _ => {}
                 }
 
-                // Construct the URL
+                // Construct the URL;
```
npmccallum, line 149: Typo

beacon/src/main.rs
```diff
@@ -135,36 +135,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
-                    Ok(true) => std::process::exit(0),
-                    Ok(false) => continue,
+                    Ok(true) => {
+                        eprintln!("[DEBUG] Action succeeded with URL {}", url);
+                        std::process::exit(0)
+                    }
+                    Ok(false) => {
+                        eprintln!("[DEBUG] Action failed with URL {}, trying next", url);
+                        continue;
+                    }
```
npmccallum, line 162: Remove the debugging.

beacon/src/main.rs
```diff
@@ -135,36 +135,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
+
                     Err(e) => eprintln!("error: {}: {}", url, e),
                 }
             }
-        });
+            Err(e) => {
+                // Resolve failed, try next item
+                eprintln!("Warning: Resolve failed: {e}");
+            }
+        }
```
npmccallum, line 171: This can all be a single line:

`Err(e) = eprintln!(...),`

beacon/src/main.rs
```diff
@@ -135,36 +135,40 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
...
     }
-
```
npmccallum, line 168: Unnecessary change.

modules/beacon/mkosi.build
```diff
@@ -0,0 +1,18 @@
...
+BEACON_DIR="${SRCDIR}/beacon"
+
+curl https://sh.rustup.rs -sSf | sh -s -- -y
+source "$HOME/.cargo/env"
+
+rustup target add x86_64-unknown-linux-musl
+
+cargo build  --release --manifest-path "${BEACON_DIR}/Cargo.toml" --target=x86_64-unknown-linux-musl
+
+cp "${BEACON_DIR}/target/x86_64-unknown-linux-musl/release/beacon" "${DESTDIR}/usr/local/bin/"
```
npmccallum, line 18: Lots of files in this commit are missing trailing newlines.

npmccallum: Rather than doing `mkdir ...; cp ...` use the `install` command which can do both at once.

npmccallum: Does this end up installing the rust toolchain into the image? If so, we should avoid that. But it isn't obvious to me.

modules/snphost/mkosi.extra/usr/local/lib/systemd/system/snphost-ok.service
```diff
@@ -1,8 +1,8 @@
 [Unit]
 Description=Run snphost ok to  make sure host is correctly set-up for SNP functionalities.
 DefaultDependencies=no
-After=network-online.target
-Wants=network-online.target
+After=network-online.target beacon-boot.service
+Wants=network-online.target beacon-boot.service
```
npmccallum, line 5: Does this actually need network-online? `beacon-boot` already requires `network-online`.

npmccallum (changes requested):
This is a big improvement.

If you're going to follow the conventional commits format, you need to use all lower case. Also, if we want to adopt that style (we don't have to), there's a GitHub action that checks PRs for conformity to that style that we should probably turn on. But we can do that after this PR merges.

DGonzalezVillal:
> This is a big improvement.
> 
> If you're going to follow the conventional commits format, you need to use all lower case. Also, if we want to adopt that style (we don't have to), there's a GitHub action that checks PRs for conformity to that style that we should probably turn on. But we can do that after this PR merges.

We actually already have a github action that checks for pr conventionality, I guess it's still missing some stuff because mine are passing, but I do appreciate the feedback.

npmccallum (approved):

PR #128, opened 2025-09-25
New build and release workflow

DGonzalezVillal:
New workflow to build the images and release them. This will make it compatible with https://github.com/AMDEPYC/snpcert/pull/90

### DIFF
--- a/modules/embed-guest-image/mkosi.build
+++ b/modules/embed-guest-image/mkosi.build
@@ -9,15 +9,17 @@ GUEST_CONFIG_DIR="${SRCDIR}/images/${GUEST_IMAGE_ID}"
 
 GUEST_IMAGE_DIR="${DESTDIR}/usr/local/lib/guest-image"
 
-if [ ! -d "$GUEST_CONFIG_DIR.efi" ]; then  
+if [ ! -f "$GUEST_CONFIG_DIR/${GUEST_IMAGE_ID}.efi" ]; then  
    echo "Guest Image not built, building image for embedding."
    #Build the guest image
-   mkosi --image-id="${GUEST_IMAGE_ID}" -C "${GUEST_CONFIG_DIR}"  build  
+   mkosi --image-id="${GUEST_IMAGE_ID}" -C "${GUEST_CONFIG_DIR}"  build
+else
+   echo "Guest Image already built, embedding found efi."
 fi
 
 if [ ! -d "$GUEST_IMAGE_DIR" ]; then  
    mkdir -p "${GUEST_IMAGE_DIR}"  
 fi
 
-cp -r "${SRCDIR}/images/${GUEST_IMAGE_ID}/${GUEST_IMAGE_ID}.efi" \
+cp -r "$GUEST_CONFIG_DIR/${GUEST_IMAGE_ID}.efi" \
    "${DESTDIR}/usr/local/lib/guest-image/guest.efi"


.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Upload host assets (custom Content-Type)
+        run: |
+          set -euo pipefail
+          # Get release id to upload via raw uploads API so we can set a custom Content-Type.
+          RELEASE_ID=$(gh api repos/$GITHUB_REPOSITORY/releases/tags/devel --jq .id)
+          shopt -s globstar nullglob
+          for efi in dist/**/host-*.efi; do
+            fname=$(basename "$efi")
+            echo "Uploading host asset $fname with custom Content-Type"
+            curl -sS -X POST \
```
npmccallum, line 65: You should definitely use `--fail` on the `curl` invocation. Otherwise you'll get silent failures.

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
+name: devel-release
```
npmccallum, line 1: As I mentioned on the other comment, this can be merged into the other .yml.

npmccallum (changes requested):

.github/workflows/build-images.yml
```diff
@@ -0,0 +1,71 @@
...
+            ${{ matrix.guest_path }}/guest-${{ matrix.distro }}-${{ matrix.release }}.efi
+          retention-days: 7
+      
+      - name: Upload host image
+        uses: actions/upload-artifact@v4
+        with:
+          name: host-${{ matrix.distro }}-${{ matrix.release }}
+          path: |
+            ${{ matrix.host_path }}/host-${{ matrix.distro }}-${{ matrix.release }}.efi
+          retention-days: 7
```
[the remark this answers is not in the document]
DGonzalezVillal: Then we don't want for the images to be built on PRs? I was thinking we could use it almost as a CI for changes in the mkosi process.

.github/workflows/build-and-release.yml
```diff
@@ -0,0 +1,127 @@
...
+
+      - name: Create release for version tag
+        run: |
+          set -euo pipefail
+          TAG="${GITHUB_REF#refs/tags/}"
+          gh release create "${TAG}" \
+            --title "Release ${TAG}" \
+            --notes "Automated build of host and guest images for ${TAG}"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
npmccallum, line 48: 1. Merge these two jobs. Tags should also delete and recreate (since we can modify tags). This means that besides the name of the job, they should be identical flows.

2. I'm not sure the `set -euo pipefail` is necessary here. There aren't any pipes.

3. These two jobs should be a step in the `build` job rather than separate jobs. Besides the obvious race condition here (that is admittedly hard to trigger), the bigger problem is that you end up deleting all the files in the release before the builds start. This means you have an empty release for a much longer duration. We want to minimize this disruption.

npmccallum (changes requested):

.github/workflows/build-and-release.yml
DGonzalezVillal: The reason this is not part of the build process is because the build process runs the matrix jobs in parallel, so each matrix job will delete and recreate the tag. This makes sure that it only happens once at push.

I can combine them, but keeping it outside the build made sense to me

PR #139, opened 2025-09-25
chore: add four targets

npmccallum:
These targets are, in order:

* boot
* exec
* report
* stop

These provide handy synchronization points for the other services.

### DIFF
--- a/modules/common/mkosi.extra/usr/local/lib/systemd/system-preset/targets.preset
+++ b/modules/common/mkosi.extra/usr/local/lib/systemd/system-preset/targets.preset
@@ -1,0 +1,1 @@
+enable stop.target
--- a/modules/common/mkosi.extra/usr/local/lib/systemd/system/boot.target
+++ b/modules/common/mkosi.extra/usr/local/lib/systemd/system/boot.target
@@ -1,0 +1,7 @@
+# Target reached when all boot services are complete
+# Boot services should use: WantedBy=boot.target
+
+[Unit]
+Description=Boot
+After=network-online.target
+Wants=network-online.target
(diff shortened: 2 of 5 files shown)


modules/common/mkosi.extra/usr/local/lib/systemd/system-preset/targets.preset
```diff
@@ -0,0 +1 @@
+enable stop.target
```
DGonzalezVillal, line 1: by just enabling stop.target, then all the other targets still get enabled because the want each other right?

modules/common/mkosi.extra/usr/local/lib/systemd/system/report.target
```diff
@@ -0,0 +1,7 @@
+# Target reached when all reporting is complete
+# Reporting services should use: WantedBy=reporting.target
```
DGonzalezVillal, line 2: Exec is running the tests and reporting is creating the certificate correct?

modules/common/mkosi.extra/usr/local/lib/systemd/system/boot.target
```diff
@@ -0,0 +1,7 @@
+# Target reached when all boot services are complete
+# Boot services should use: WantedBy=boot.target
```
DGonzalezVillal, line 2: Would beacon boot be a part of the boot target?

DGonzalezVillal (commented):
Just some questions to better understand how you want to use these targets. 

The bot review should probably be addressed still before approving. 

Implementing these targets is going to be a different PR then?

PR #141, opened 2025-09-26
Configure AMDSEV compatible OVMF path based on OS

LakshmiSaiHarika:
Configure AMDSEV compatible OVMF path based on the OVMF package provisioned in the OS images on host platform to calculate guest measurement and launch SNP guest measurement.


DGonzalezVillal (approved):
This solution works.  Just a reminder that in this solution OVMF_PATH will get assigned the value of the last item in the loop, so [-z OVMF_PATH] would evaluate to false, even if OVMF_PATH is not a "valid" path.  If we want to avoid this we could do something like this instead:

```bash
OVMF_PATH=""
for p in /usr/share/ovmf/OVMF.amdsev.fd /usr/share/edk2/ovmf/OVMF.amdsev.fd; do
    if [ -f "$p" ]; then
        OVMF_PATH="$p"
        break
    fi
done

if [ -z "$OVMF_PATH" ] || [ ! -f "$OVMF_PATH" ]; then
    echo "AMDSEV compatible OVMF is not present, can't calculate measurement" >&2
    exit 1
fi
```

Where OVMF_PATH only gets a value assigned to it if the the correct path is found.

DGonzalezVillal (approved):

PR #145, opened 2025-09-30
fix: resolve network initialization issue

DGonzalezVillal:
The network preset lacked a numeric prefix, causing ti to be ignored. As a result, tests with soft network requirements could start before the network was fully up, leading to boot failures and network-related errors later in testing.


LakshmiSaiHarika (approved):
LGTM

PR #168, opened 2025-10-16
feat: support reporting the milestone

amd-aliem:
Add the milestone upon test success.

Unsure if this is the best way to parse the certificate file. Perhaps it would be better to grep for success - but to do that may need to add a line checking [tests passed]/[total tests] that we can grep.

### DIFF
--- a/images/guest-debian-forky/mkosi.conf
+++ b/images/guest-debian-forky/mkosi.conf
@@ -1,0 +1,24 @@
+[Include]
+Include=../../modules/load-kernel-modules
+Include=../../modules/guest
+
+[Distribution]
+Distribution=debian
+Release=forky
+
+[Content]
+Packages=
+	linux-image-amd64
+	systemd
+	systemd-boot
+	systemd-boot-efi
+	systemd-resolved
+	plymouth
+	login.defs
+	login
+	locales
+	systemd-journal-remote
+	openssl
+	ca-certificates
+	jq
+	xxd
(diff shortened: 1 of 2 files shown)


modules/beacon/mkosi.extra/usr/local/lib/scripts/beacon-report.sh
```diff
@@ -29,7 +29,12 @@ for sev_version in "${SEV_VERSIONS[@]}"; do
   SEV_CERT_FILE="${HOME:-/root}/sev_certificate_v${sev_version}.txt"
 
   # Call beacon
-  beacon report --title "$SEV_TITLE" --body "$SEV_CERT_FILE" --label "certificate" --label "${OS_LABEL}"
+  if [ -e "${SEV_CERT_FILE}" ] && [ -z "$(grep "❌" "${SEV_CERT_FILE}")" ]; then
+    # Add milestone if no errors encountered
```
npmccallum: So, that''s not how the certificate generation is supposed to work. We might need broader changes here if it doesn't meet the design criteria. Here's how it is supposed to work:

1. Every test has an associated version number.
2. We run all tests.
3. We certify a distro with the newest version number that has no failed tests in equal or later versions.

For example, if a distro runs the following tests:
* foo (v3.1-0) PASS
* bar (v3.0-8) FAIL
* baz (v3.0-4) PASS

We would certify the distro only for `v3.0-4`. The version that applies to each test is specify as extra data in the systemd unit file.

npmccallum (changes requested):

DGonzalezVillal:
@amd-aliem this looks good, but it seems you need to rebase your branch

PR #221, opened 2025-12-16
docs: dispatch proxy handling

amd-aliem:
Break out GH permissions into another section and then create a section for proxy handling. Handling https://github.com/AMDEPYC/dispatch/issues/17.

### DIFF
--- a/docs/how-to-generate-certs.md
+++ b/docs/how-to-generate-certs.md
@@ -36,7 +36,8 @@ References:
 
 ### 2. Download `AMDEPYC/dispatch`
 
-Download the [latest dispatch binaries](https://github.com/AMDEPYC/dispatch/releases) through your browser or through `gh`:
+Download the [latest dispatch binaries](https://github.com/AMDEPYC/dispatch/releases)
+through your browser or through `gh`:
 ```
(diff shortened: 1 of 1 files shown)


docs/how-to-generate-certs.md
```diff
@@ -91,15 +66,25 @@ You can optionally add a filter at the end if you want to run on a subset of ima
...
+You can also fork the `AMDEPYC/sev-certify` repository to build the images and
+have certification results posted in that repository's issues - see
+[Forking sev-certify](#forking-amdepycsev-certify-for-certification-testing) for
+more information.
 ```
 ./dispatch-linux-x86_64 --owner <your org/username> --repo sev-certify --tag devel
 ```
 
-If you run the first command with no filter specified, you should see the following screen:  
+If you run the first command with no filter specified, you should see the
```
DGonzalezVillal:
```suggestion
If you run the first command with no filter specified, you see all of the current images being hosted in the disptach release: 
```

I'm not sure if this is the perfect wording, but I wouldn't tell them they are getting the exact same screen, since it could change every time we change the available images. For example, it is already outdated

DGonzalezVillal (changes requested):
Some nitpicks

PR #236, opened 2026-03-19
Target refactor

DGonzalezVillal:
# Refactor: Introduce staged systemd target pipeline

## Overview

This PR restructures the execution flow into a staged pipeline using systemd targets and barrier services. The goal is to make the workflow easier to understand, extend, and maintain as the project grows.

Instead of relying on loosely coupled services with implicit ordering, the pipeline is now explicitly modeled as a sequence of targets:

Host:
boot → system → launch → test → report → stop

Guest:
boot → system → test → report → stop

Each stage is responsible for a well-defined portion of the workflow and is enforced through a corresponding barrier service.

---

### Targets

Each stage is represented by a systemd target:

- `boot.target` – waits for required system services
- `system.target` – performs system validation checks
- `launch.target` – launches and verifies the guest
- `test.target` – executes test workflows
- `report.target` – collects results and generates certificates
- `stop.target` – finalizes execution and reboots

Targets are chained using:

```ini
Requires=<previous-stage>.target
After=<previous-stage>.target
```
This allows the pipeline to be driven entirely by systemd’s dependency graph.

### Barrier Services

Each stage includes a `*-done.service` that acts as a synchronization point:

- Pulls in all services for that stage
- Waits for them to complete
- Marks the stage as finished

Example:

```ini
[Unit]
Requires=service-a service-b
After=service-a service-b

[Service]
Type=oneshot
ExecStart=/usr/bin/true
RemainAfterExit=yes
```

Barrier services:
- enforce stage completion
- prevent race conditions
- replace the need for `OnSuccess=` / `OnFailure=` transitions

---

### Dependency Model

- **Targets define structure**
- **Barrier services enforce correctness**
- **Stage services perform actual work**

Dependency usage:

- `Requires=` → strict stage dependencies
- `Wants=` → non-critical services (e.g., networking, logging)
- `After=` → ordering only

---

### 1. Pipeline Reorganization

- Introduced explicit staged targets
- Removed implicit and hard-to-follow service dependencies
- Eliminated `OnSuccess=` / `OnFailure=` transitions in favor of graph-driven execution

### 2. Separation of Concerns

- System validation (`snphost`, `snpguest`) moved to `system.target`
- Test execution isolated to `test.target`
- Reporting centralized in `report.target`

### 3. Reporting Improvements

- We are still hard coding values for our most recent certification version. We were hard coding the old value of 3.0-0, but it's been updated to 3.0.0-0 so it presents the correct value. The reporting/testing stages will be modified in a future PR to remove the need for this hard coding. 

### 5. Module Consolidation

- Centralized build modules into host, guest, and common categories
- Updated module logic to self-determine applicability
- Removed need for selective module inclusion per image
- Each stage it's own folder under modules for a neater project organization.

### 6. Host / Guest Separation

- Introduced target differences where necessary (e.g., no `launch.target` on guest)
- Maintained symmetry while allowing role-specific behavior

---

## Behavior Changes

- Execution is now driven by `stop.target` as the entry point
- The entire pipeline is resolved via systemd dependencies
- Stage progression no longer relies on explicit transitions
- Failures propagate naturally through the dependency graph
- Reporting always executes before shutdown/reboot

---


modules/system/host/beacon-boot/mkosi.extra/usr/local/lib/systemd/system/beacon-boot.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
```
npmccallum: The old version of this service had `DefaultDependencies=no` — same for several others that got moved (beacon-report, boot-successful, launch-guest, etc.). Was that a deliberate change? With the default on, systemd will add implicit `Requires=sysinit.target` and `Conflicts=shutdown.target`. Just curious if you've tested on a UKI boot to make sure that doesn't cause any ordering surprises.

modules/build/guest/targets/mkosi.extra/usr/local/lib/systemd/system/system.target
```diff
@@ -0,0 +1,11 @@
+# Target reached when the system checks have completed.
+# Job services should use: WantedBy=system.target
+
+[Unit]
+Description=System stage
+
+After=boot.target
+Requires=boot.target
```
npmccallum: The old targets used `Wants` (soft dependency) here — was the switch to `Requires` intentional? It'll make the chain fail-fast, which could be what you want for certification. Just want to make sure a single flaky service in an early stage won't block the whole run unexpectedly.

npmccallum (commented):
Really like the direction here — organizing by lifecycle stage makes the execution flow much easier to follow at a glance, and the barrier pattern with the `*-done.service` units is clean. The merged `launch-and-verify.sh` is a nice improvement too.

Left a few questions inline.

LakshmiSaiHarika (approved):
Looks good!!!

amd-aliem (approved):

markg-github:
I used claude code to get this review started. I used a fairly generic skill, but I also told claude to focus on:

1) changes to amount of dependence on systemd, in general, and specifically for sequencing of operations
versus, for example, checking/polling logs as a synchronization/sequencing mechanism. Theoretically, we only absolutely need systemd for automation purposes, ie, automatically running tests (in the form of systemd services) after boot
2) changes to "types" of systemd units and necessity of any such changes
3) what are the new systemd units (intro'd by this patch/PR)? 
4) what are/does the author mean by "barrier services" and what are the barrier services?


Item 1) above goes to the question of what we would do if systemd wasn't available (to hosts), as in Windows/Hyper-V?

## Review: Target refactor

*AI-assisted review focused on systemd architecture, barrier services, and sequencing. Human reviewer: @mgentry*

---

### Overall

The staged pipeline design is a clear improvement over the previous flat structure. Explicit stage targets, barrier services, and removal of `OnSuccess=`/`OnFailure=` transitions make the execution model much easier to reason about. The barrier service pattern (`Type=oneshot`, `ExecStart=/usr/bin/true`, `RemainAfterExit=yes`, `DefaultDependencies=no`) is correctly implemented and well-suited to this use case.

Two items warrant discussion before merge, and one question needs confirmation.

---

### Discussion: `Wants=` vs `Requires=` on barrier services

Every stage target uses `Wants=` (not `Requires=`) to pull in its barrier:

```ini
# e.g. launch.target
Wants=launch-done.service
After=launch-done.service
```

The consequence: if a barrier service fails (because a stage service failed), the target still reaches its active state. The next stage's `Requires=<prev>.target` is satisfied, and the pipeline continues.

Concrete failure path if `launch-guest.service` fails:

```
launch-guest.service FAILS
  → launch-done.service FAILS (Requires= launch-guest)
    → launch.target stays ACTIVE (only Wants= the barrier)
      → test.target STARTS
        → report.target STARTS → collects logs, generates cert
          → stop.target → reboot
```

**Question for author:** Is this intentional? If so, the rationale would be: always run to completion (including reboot) regardless of intermediate failures, so the machine is never left in a half-run state. That's a defensible design for an automated system. If not intentional, `Requires=launch-done.service` (etc.) would halt the pipeline at the failed stage.

The PR description says "failures propagate naturally through the dependency graph" — with `Wants=`, they don't propagate at the target boundary, so please clarify the intended failure behavior.

---

### Issue: Host `test-done.service` has no dependencies

```ini
# modules/test/host/test-done/...
[Unit]
Description=Barrier that triggers test services
DefaultDependencies=no
# No Requires=, no After= sub-services

[Service]
Type=oneshot
ExecStart=/usr/bin/true
RemainAfterExit=yes

[Install]
WantedBy=test.target
```

Unlike the guest `test-done.service` (which properly `Requires=attestation-result.service attestation-workflow.service`), the host barrier has no dependencies and immediately succeeds. This means `test.target` on the host completes essentially instantly, and `report.target` can start before guest tests are finished.

If there are intentionally no host-side test services in this PR (i.e., the host test stage is a structural placeholder for future work), that's fine — but it should be noted. If host tests are expected to run in this stage, the barrier needs its dependencies filled in.

---

### Observation: Host-guest synchronization is still polling-based

The refactor cleanly handles intra-host and intra-guest ordering via the dependency graph. However, the host waiting for guest test completion is still handled by `display-guest-logs.sh` polling the remote journal:

```bash
TIMEOUT=60
INTERVAL=1
ELAPSED=0
units=("snpguest-ok.service" "attestation-workflow.service")
```

This is the existing mechanism and not a regression — just noting that the cross-machine synchronization boundary remains outside the systemd graph. Worth tracking as future work if the goal is to eliminate polling entirely.

---

### Minor

- `tests-finished.service` description still contains a grammar error: "it's tests" → "its tests"
- Version string `3.0-0` → `3.0.0-0` confirmed intentional per PR description — no concern.

---

### Clean areas (safe to skim)

- All 14 image `mkosi.conf` files — mechanical path updates only
- CI `build-and-release.yml` — 3 path updates matching module reorganization  
- Module renames/moves throughout `modules/` — confirmed as renames by diff similarity scores
- Certificate generator Python — moved files plus intentional version string fix

PR #242, opened 2026-05-13
Add a derived keys test

### LINKED ISSUE #105: key derivation test

markg-github:
The test invokes snpguest repeatedly, especially the key subcommand, to test the SEV-SNP derived keys feature. The test mostly verifies that the derived keys are different when they're supposed to be.

More complete testing could be achieved by including an ID block. Including an ID block is completely independent and I plan a PR for it soon.

I think AI did a decent job at maintaining "certificate consistency", but I'm not sure about the versioning.

Closes #105 and https://amd.atlassian.net/browse/DEAEOCC-304 (Jira).


amd-aliem:
@markg-github I had added some certificate versioning logic in https://github.com/AMDEPYC/sev-certify/pull/241 (it still needs to get reviewed), but I think it might be good to discuss what we want to keep and how we'll merge both. For that one I was intending the next few tests to be under v3.0.0-1.

Do you have some passing certificate examples you can link here? It should have a milestone assigned if it reaches some certificate level. I looked at your recent ones in your repository and they appear to have issues. One of the issues looked familiar, I had a fix for snpguest ok running too early sometimes. But not sure if the others were expected. Note that you'll need to create the milestone in your fork for it to be assigned if you haven't yet (c3.0.0-0, and later when we enable it, c3.0.0-1)

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
```diff
@@ -0,0 +1,870 @@
...
+      bits  7:0  - Boot Loader SVN
+      bits 15:8  - TEE SVN
+      bits 47:16 - Reserved (zero)
+      bits 55:48 - SNP firmware SVN
+      bits 63:56 - Microcode SVN
+    """
+    boot_loader: int = 0
+    tee: int = 0
+    snp: int = 0
+    microcode: int = 0
```
DGonzalezVillal: This will be different according to generation:

This matches the Milan and Genoa TCB spec. We need to extend to the Turin definition. Venice will be a completely different look as well.

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
```diff
@@ -0,0 +1,870 @@
...
+            ((self.snp & 0xFF) << 48) |
+            ((self.microcode & 0xFF) << 56)
+        )
+
+    def __str__(self) -> str:
+        return (f"bl=0x{self.boot_loader:02x} tee=0x{self.tee:02x} "
+                f"snp=0x{self.snp:02x} mc=0x{self.microcode:02x}")
+
+
+EXPECTED_REPORT_VERSION = 2  # ATTESTATION_REPORT schema version this test was written for
```
DGonzalezVillal: Is there a particular reason you need to be aware of the report version?

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
```diff
@@ -0,0 +1,870 @@
...
+        ]:
+            pattern = rf'{re.escape(section_name)}\s*:?(.*?)(?={boundary}|\Z)'
+            m = re.search(pattern, display_output, re.DOTALL | re.IGNORECASE)
+            if m:
+                setattr(info, attr, parse_tcb_section(m.group(1)))
+
+        return info
+    except Exception as e:
+        print(f"WARNING: Failed to parse report info: {e}", file=sys.stderr)
+        return None
```
DGonzalezVillal: I'm wondering, I have not implemented a parser in python, but maybe if it would be easier to find the info you're looking for by looking at the bit ranges rather than parsing the display format of snpguest?

like read bytes in and parse the specific locations.

Quick AI gave me this:

```python

import struct

def extract_tcb_values(report_filepath):
    # Dictionary mapping the TCB names to their byte offset in the specification
    # Each TCB version field is 64 bits (8 bytes) long
    tcb_offsets = {
        "CURRENT_TCB": 0x38,
        "REPORTED_TCB": 0x180,
        "COMMITTED_TCB": 0x1E0,
        "LAUNCH_TCB": 0x1F0
    }
    
    tcb_results = {}
    
    with open(report_filepath, 'rb') as f:
        for name, offset in tcb_offsets.items():
            # Seek directly to the byte location
            f.seek(offset)
            
            # Read exactly 8 bytes (64 bits)
            byte_data = f.read(8)
            
            if len(byte_data) < 8:
                tcb_results[name] = "Error: Out of bounds / File too short"
                continue
            
            # Unpack as an unsigned 64-bit integer (<Q for Little-Endian, >Q for Big-Endian)
            # x86 platform hardware structures typically use Little-Endian format
            tcb_value = struct.unpack('<Q', byte_data)[0]
            tcb_results[name] = tcb_value
            
    return tcb_results

def decode_tcb_components(tcb_int):
    # Convert the 64-bit integer back to 8 raw bytes
    tcb_bytes = struct.pack('<Q', tcb_int)
    
    # AMD SEV-SNP maps these to individual security version numbers (SVNs)
    return {
        "Microcode SVN": tcb_bytes[0],
        "SNP SVN":       tcb_bytes[1],
        "TEE SVN":       tcb_bytes[2],
        "Bootloader SVN":tcb_bytes[3],
        # Bytes 4-7 are typically reserved or specific platform SVNs
        "Raw Bytes":     list(tcb_bytes)
    }
```

Not mandatory but I feel this type of data works easier with bytes, but also python is not created to work with bytes. So both approaches are fine with me.

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
```diff
@@ -0,0 +1,870 @@
...
+
+    dprint(report_text)
+
+    report_info = parse_report_info(report_text)
+    if report_info:
+        if report_info.version is not None:
+            version_note = (
+                "" if report_info.version == EXPECTED_REPORT_VERSION
+                else f" *** UNEXPECTED (expected {EXPECTED_REPORT_VERSION}) —"
+                     f" TCB layout assumptions may not apply ***"
```
DGonzalezVillal: Oh I see your logic is that the TCB changes according to your report version. 

The TCB really changes according to your CPU as I noted before. The only way of being able to tell what CPU you're running on (from a guest perspective) is through the attestation report family/model/stepping fields, and those are only present on reports v3+. I think we can fix up this TCB logic up a bit on the new harness, it should be easier to extract some of this information.

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
```diff
@@ -0,0 +1,870 @@
...
+        dprint(f"  VMPL1 Key: 0x{key_vmpl1_hex}")
+        print("✓ PASS: VMPL0 and VMPL1 keys differ (proper isolation)")
+        return True
+    else:
+        print("✗ FAIL: VMPL0 and VMPL1 keys are identical", file=sys.stderr)
+        return False
+
+
+def test_root_key_difference() -> bool:
+    """Test that different root keys (VCEK vs VMRK) produce different keys."""
```
DGonzalezVillal: This is just out of my own curiosity, but did you find out if there is a way of generating, or finding the vmrk? I know we can request vcek from kds, but I have no idea where the vmrk lives, or if there is ever a situation where you want to generate it.

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
```diff
@@ -0,0 +1,673 @@
...
+    key_vck_file  = KEY_DERIVATION_DIR / "vck_key.bin"
```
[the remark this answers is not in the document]
DGonzalezVillal: @markg-github I know that you meant vcek ignore this comment.

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
```diff
@@ -0,0 +1,870 @@
...
+    max_svn = report_info.guest_svn if report_info is not None else 0
+    print(f"  Guest SVN upper bound: {max_svn}")
+
+    svn_values = list(range(0, max_svn + 1))
+
+    if len(svn_values) < 2:
+        print("  Only one valid SVN value (0); sensitivity cannot be tested.")
+        print("  (Expected when guest was launched without an ID block.)")
+        print("✓ PASS: N/A (single valid value)")
+        return True
```
DGonzalezVillal: If guest svn is 0, could you at least test that 1 fails? I don't know if that's how it works. But maybe if svn is 0, then try to derive with 1 and expect a failure. And then only test ranges if svn is > 0.

DGonzalezVillal (commented):
You went above and beyond! I think this really covers all of the derived key functionality testing, covered pretty much most test cases. 

I added some comment on you tcb derivation logic, tcb is derived from chip and not from report version. 

Also just an idea on how to maybe parse the contents of the report. I know historically we've done it by printing the report and the parsing the contents, but I wonder if for us trying to handle bytes could be cleaner. 

I also think that the way you have it set up could lend itself well to migrate to the new test harness. 

For some of the tests you're generating a large amount of keys, that's the one thing that worries me since the new harness doesn't really have a way of running functions on guest, so we might need to brainstorm how to migrate the size of your test into the harness.

modules/build/guest/mkosi.conf
```diff
@@ -10,3 +10,5 @@ Include=../../stop/guest
 
 [Content]
 KernelCommandLine=console=ttyS0
+Packages=
+	python3
```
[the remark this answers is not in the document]
amd-aliem: Don't think it overwrites like copilot says, but at this point not needed anymore since we added python3 for the sev_verify harness already (@DGonzalezVillal's vsock change), so guests already have python3 (unless you're trying to deduplicate, but then I think that can be its own PR?)

modules/report/host/display-guest-logs/mkosi.extra/usr/local/lib/scripts/display-guest-logs.sh
```diff
@@ -7,7 +7,7 @@ TIMEOUT=60
 INTERVAL=1
```
amd-aliem: this file's been removed in https://github.com/AMDEPYC/sev-certify/commit/e421a9f0490d22f173c2206543bbb701935e4489

modules/report/host/sev-certificate-generator/mkosi.extra/usr/local/lib/scripts/generate_sev_certificate/service/service.py
```diff
@@ -41,6 +41,10 @@ def get_service_description(self, service, platform):
```
amd-aliem: this file's been removed in https://github.com/AMDEPYC/sev-certify/commit/e421a9f0490d22f173c2206543bbb701935e4489

modules/report/host/sev-certificate-generator/mkosi.extra/usr/local/lib/scripts/generate_sev_certificate/sev_certificate/sev_certificate_version_3_0_0_0.py
```diff
@@ -109,6 +109,46 @@ def get_snp_guest_attestation_summary(self):
 
         return snpguest_attestation_summary
 
+    def get_key_derivation_summary(self):
```
amd-aliem: this file's been removed in https://github.com/AMDEPYC/sev-certify/commit/e421a9f0490d22f173c2206543bbb701935e4489

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
```diff
@@ -0,0 +1,871 @@
+#!/usr/bin/env python3
+"""
```
amd-aliem: imo this script should be co-located with the module in sev_verify

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/systemd/system/key-derivation.service
```diff
@@ -0,0 +1,18 @@
+[Unit]
+Description=Run SNP Key Derivation Tests after boot
```
amd-aliem: why is this a service? I see you're already using the .for_guest() step type, which is what I think would launch it. I think this file should be removed unless I'm missing something?

modules/test/guest/test-done/mkosi.extra/usr/local/lib/systemd/system/test-done.service
```diff
@@ -2,8 +2,8 @@
 Description=Barrier that triggers test services
```
amd-aliem, line 2: removed by https://github.com/AMDEPYC/sev-certify/commit/e421a9f0490d22f173c2206543bbb701935e4489

modules/test/guest/mkosi.conf
```diff
@@ -1,4 +1,5 @@
 [Include]
```
amd-aliem, line 1: removed by https://github.com/AMDEPYC/sev-certify/commit/e421a9f0490d22f173c2206543bbb701935e4489

sev_verify/cert_tests/c3_0/manifest.toml
```diff
@@ -19,3 +19,12 @@ description = "Test SNP_CONFIG and SNP_COMMIT host commands"
...
 level = "3.0.0-1"
+
+# ── Level 3.0.0-2 ───────────────────────────────────────────
+
+[[tests]]
+name = "key-derivation"
+description = "Launch SEV-SNP guest and run key derivation tests"
+module = "cert_tests.c3_0.c3_0_0_2.key_derivation_test"
+scope = "mixed"
+level = "3.0.0-2"
```
amd-aliem: I believe we were having this under 3.0.0-1 (released together with snp config commit?), not one test per -X level

sev_verify/cert_tests/common/snp_ok.py
```diff
@@ -47,7 +47,7 @@ def steps() -> list[BaseStep]:
     return [
         Step.for_host(
             name="snphost ok",
-            type="required",
+            type="info",
             command="snphost ok",
         ),
```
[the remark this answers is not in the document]
amd-aliem: Not sure I agree with changing this to `info`... or at least, a fail here shouldn't grant a milestone/pass even if everything else passes.

You added this b/c your machine was failing the snphost alias check right? I briefly asked @DGonzalezVillal and @amd-kkin if we should move the alias check into a --mitigations flag or something (along with other cve checks in the future). Maybe that would help here and more reason to keep it required.

amd-aliem (changes requested):
Reviewed only for structure, it currently has a lot of extra files from the legacy systemd modules - if you could please rebase against main and clean that up would make it a lot shorter.

Will be doing another review on the test script and content tomorrow!

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
amd-aliem: thinking about this more, it may be harder than I thought since we don't have a 'file push' for vsock yet. I'm ok with keeping it here for now - but really would like to discuss the convention for future tests and potentially move this script to sev_verify in the future.

Reasons I would prefer to NOT add it to the guest image:
- harder to test changes (whole image build for any change to this script)
- image will get larger unnecessarily, esp if we allow multiple VM launches. Bring up a VM to test 4.0 will have extra data from all previous tests.

@DGonzalezVillal maybe a new step type for "run this script on the VM" that bundles together pushing a script and then executing it?

markg-github:
I changed to use the generic guest image. The message from the one new commit:

feat: move key derivation test logic to host, use vsock for guest commands

    Replace the guest-side Python script approach with individual snpguest
    key commands over vsock, pulling derived keys to the host for comparison.
    All test logic (determinism, VMPL isolation, root key difference, SVN/TCB
    bound enforcement, GFS sensitivity and field mixing) now lives in the
    sev_verify test module as callable steps.

    Use callable steps that loop over snpguest key via
    run_guest_command/fetch_guest_file_bytes while the VM is running. SVN
    and TCB sweeps are fully dynamic based on the runtime attestation report.

    TCB above-bound tests use committed+1..committed+3 per component from
    the runtime report, capped at 0xFF to avoid 8-bit overflow.

    Remove key-derivation guest module. The guest-side Python script and systemd
    service are no longer needed and snpguest is available in the generic guest
    image.

modules/build/guest/mkosi.conf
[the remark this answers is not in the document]
DGonzalezVillal: I think if it's named differently it might cause issues.

sev_verify/cert_tests/c3_0/c3_0_0_2/key_derivation_test.py
```diff
@@ -0,0 +1,493 @@
...
+    snp: int = 0
+    microcode: int = 0
+
+    def to_u64(self) -> int:
+        return (
+            (self.boot_loader & 0xFF) |
+            ((self.tee & 0xFF) << 8) |
+            ((self.snp & 0xFF) << 48) |
+            ((self.microcode & 0xFF) << 56)
+        )
```
DGonzalezVillal, line 68: Your TCB generation still needs updating for Turin version (I'll allow no Venice for now)

sev_verify/cert_tests/common/snp_ok.py
```diff
@@ -47,7 +47,7 @@ def steps() -> list[BaseStep]:
     return [
         Step.for_host(
             name="snphost ok",
-            type="required",
+            type="info",
```
DGonzalezVillal:
```suggestion
            type="required",
```

sev_verify/cert_tests/common/snp_ok.py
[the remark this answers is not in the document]
DGonzalezVillal: I asked mark to move it back to required, but do you think info is fine for snphost ok check? I think we need to introduce the fix to snphost ok, but keep this particular step as required.

DGonzalezVillal (changes requested):
Need to extend the TCB definition for Turin vs Genoa and consider Venice one will look different as well. If need be we can chat on ways to do this.

sev_verify/cert_tests/common/snp_ok.py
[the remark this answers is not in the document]
amd-aliem: Yeah I would prefer to keep it as required for now

sev_verify/cert_tests/c3_0/c3_0_0_2/key_derivation_test.py
```diff
@@ -0,0 +1,493 @@
...
+        return StepHandlerResult(exit_code=1, stderr=err1 or err2)
+    k1 = _read_key(ctx.artifact_dir / "det_key1.bin")
+    k2 = _read_key(ctx.artifact_dir / "det_key2.bin")
+    if k1 == k2:
+        return StepHandlerResult(exit_code=0, stdout="Keys match (deterministic)")
+    return StepHandlerResult(exit_code=1, stderr="Keys differ — derivation is not deterministic")
+
+
+def test_vmpl_isolation(ctx: StepContext) -> StepHandlerResult:
+    ok0, err0 = _derive_key(ctx, "vmpl0_key.bin", vmpl=0)
```
amd-aliem: For forwards compatibility should we NOT hard-code vmpl=0? In the fuure possibly distros may have svsm enabled by default so VMs would be on vmpl 1 right? Could check current level first and then increment by 1 perhaps. Also ok with leaving this for whenever we run into it.

markg-github:
Add so-called cross-CVM (sub) test: launch a CVM - save a derived key - stop the CVM - launch a CVM - verify ability to generate the same key.

This version of the test was in a 2nd, draft PR (#285) --> decided no need to split up the test across two PRs and to close the 2nd PR.

sev_verify/cert_tests/common/snp_ok.py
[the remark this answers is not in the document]
markg-github: I'm fine with changing it back to required and I did so. (I also updated my system's TCB enough to avoid the failure.)

fwiw, it seems that the alias check issue/mitigation at play here is different, if not entirely unique, in that the mitigation isn't effective if loaded at runtime. That said, it's not intuitive to me that missing "reboot required" mitigations like this should be considered failures while missing "runtime okay" mitigations aren't.

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
markg-github: I addressed by adding the following to the main test file, key_derivation_test.py:

  from sev_verify.guest_vsock import GuestCommandError, fetch_guest_file_bytes, run_guest_command

and using GuestCommandError, fetch_guest_file_bytes and run_guest_command directly.

To the point raised by @amd-aliem and @DGonzalezVillal , I agree that it would be better if sev_verify itself supported "dynamic tests" like this better.

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/systemd/system/key-derivation.service
markg-github: fixed

sev_verify/cert_tests/c3_0/manifest.toml
markg-github: No problem. I'll change.

sev_verify/cert_tests/c3_0/c3_0_0_2/key_derivation_test.py
markg-github: Good point. 

cf9fe4b3c33a

modules/build/guest/mkosi.conf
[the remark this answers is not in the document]
markg-github: I'll remove and update the PR again after testing.

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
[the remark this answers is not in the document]
markg-github: I believe I saw "VCK" somewhere and it may me wonder about something that I was already curious about: in the context of derived keys, VCEK, as an asymmetric key (pair), shouldn't be relevant. What's relevant is the derivation chain from fuses (PUFs?) to the derived keys and I thought VCK might be the name of something in this chain.

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
markg-github: ACK

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
markg-github: You mean since the code doesn't parse the report directly, per your earlier comment?

In general, I would say any type of parsing should condition on a (schema/struct) version whenever possible.

If you're asking why I need to read the report in the 1st place, it's to learn what the derived keys valid (and invalid) inputs are.

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
markg-github: I remembered this comment, but thought it was on my ID block PR. No problem, they should both use the same approach. I'll either split out the binary parser (from ID block PR) into a new PR or if ID block merges, rebase and use the binary parser. I do NOT want to "stack" any PRs. :)

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
markg-github: Not sure. It's migration-related, right? If I had to guess, I'd say a more or less dedicated FW API for it...?

modules/test/guest/key-derivation/mkosi.extra/usr/local/lib/scripts/snpguest_key_derivation.py
markg-github: I tried to generalize your idea here, @DGonzalezVillal , in updated version(s) of the PR.

modules/build/guest/mkosi.conf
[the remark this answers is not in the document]
markg-github: eaeda0690

sev_verify/cert_tests/c3_0/manifest.toml
markg-github: cf9fe4b3c33a

sev_verify/cert_tests/common/snp_ok.py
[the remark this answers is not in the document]
markg-github: Attestation report parsing is in flux. I'm now leaning toward at least being consistent until snpguest provides machine-readable (JSON) output (the agreed upon direction). Here, this likely means waiting for #247 to merge and then using its parser. The snpguest changes are presumably further out.

PR #245, opened 2026-05-20
feat: sev_verify test harness base

amd-aliem:
First pass at a host-side testing harness for SEV certification. Reads TOML manifests that declare which tests to run, imports per-test Python modules that define executable steps, and orchestrates execution across host and guest environments.

- models.py and snphost_ok.py contents are temporary, likely will change as we start wiring steps together. Mostly to show an example of how a test could be defined.
- test harness allows running from a compliant sev-enabled host, but will eventually also be pulled into host images for dispatch certification testing.

Example output:
```
$ sudo python3 -m sev_verify images/guest-ubuntu-25.04/guest-ubuntu-25.04.efi
   Guest: images/guest-ubuntu-25.04/guest-ubuntu-25.04.efi

── Prerequisites ───────────────────────────────────────────────────────────────

   snphost-ok               Verify SEV-SNP is enabled and functional on the host
   ├─ snphost-ok ··················································· 30ms [PASS]
   └─ snphost-show-guests ·········································· 11ms [PASS]
   result:                                                     2/2 passed [PASS]

   Prerequisites:                                              1/1 passed [PASS]

── Certification 3.0 ───────────────────────────────────────────────────────────
   SEV 3.0 Tests (AMD EPYC 7003+)

   ── 3.0.0-0 ──
   vm-launch-attest        Launch an SEV-SNP guest and verify attestation report
   └─ vm-launch-dummy ··············································· 1ms [PASS]
   result:                                                     1/1 passed [PASS]

   ── 3.0.0-1 ──
   snphost-config-commit            Test SNP_CONFIG and SNP_COMMIT host commands
   ├─ show-tcb ····················································· 10ms [PASS]
   ├─ config-set-lower ············································ 126ms [PASS]
   ├─ config-reset ················································· 70ms [PASS]
   ├─ config-set-mask-chip-id ······································ 16ms [PASS]
   ├─ config-set-mask-chip-key ····································· 15ms [PASS]
   ├─ config-set-mask-both ········································· 16ms [PASS]
   ├─ config-reset-masks ··········································· 71ms [PASS]
   └─ commit ························································ 5ms [PASS]
   result:                                                     8/8 passed [PASS]

   Certification 3.0:                                          2/2 passed [PASS]

── Summary ─────────────────────────────────────────────────────────────────────
   Certification 3.0: (level 3.0.0-1)                                     [PASS]

```

### DIFF
--- a/sev_verify/cert_tests/cert_3_0/manifest.toml
+++ b/sev_verify/cert_tests/cert_3_0/manifest.toml
@@ -1,2 +1,0 @@
-version = "3.0"
-description = "SEV 3.0 Tests (AMD EPYC 7003+) - Current Level 3.0.0-0"
--- a/sev_verify/cert_tests/cert_3_0/manifest.toml
+++ b/sev_verify/cert_tests/cert_3_0/manifest.toml
@@ -1,0 +1,2 @@
+version = "3.0"
+description = "SEV 3.0 Tests (AMD EPYC 7003+) - Current Level 3.0.0-0"
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,5 @@ images/*/*.initrd
 images/*/*.vmlinuz
 images/*/guest-*
 images/*/host-*
+
+results/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,9 @@
+__pycache__/
+*.pyc
+*.egg-info/
+dist/
+build/
+
 **/target/
 /*/tools/
 /*/image
--- a/sev_verify/README.md
+++ b/sev_verify/README.md
@@ -1,0 +1,46 @@
+# sev_verify
+
+Host-side testing harness for SEV-SNP certification. Reads TOML manifests that declare which tests to run, imports per-test Python modules that define executable steps, and orchestrates execution across host and guest environments.
+
+## Usage
+
+```bash
+# Run a specific certification level
+python3 -m sev_verify /path/to/guest.efi -v 3.0
+
+# Run multiple levels
+python3 -m sev_verify /path/to/guest.efi -v 3.0 -v 3.1
+
+# Run all certifications found in cert_tests/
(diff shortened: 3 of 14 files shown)


sev_verify/cert_tests/cert_3_0/manifest.toml
DGonzalezVillal, line 1: Are you/we planning to only maintain one certificate definition per generation?

Or what is your idea for the certificate definition structure?

I was looking at your command-line arguments and noticed that we would run certificates per CPU generation. I do like that approach.

My original intent was to go much more granular, something like:

-v 3.0.0-0 -v 3.1.2-0 -v 3.0.1-2

But I feel like that would require users to type too much if they want to run more than one test.

We’ll keep it as you proposed, but I’m curious how you envision the full manifest evolving as we add new versions across the different certificate generations.

DGonzalezVillal: Or I guess, now that I read a little more, I think I would need an example on what a full manifest would look like.

DGonzalezVillal (commented):
I like the  manifest idea. I think I'm still a little unclear on what the manifest would look like and how the project would flow. Would the full cert-definitions be inside the manifest? Like the cert, test and steps would be written there?

sev_verify/cert_tests/cert_3_0/manifest.toml
amd-aliem: > My original intent was to go much more granular, something like:
>     -v 3.0.0-0 -v 3.1.2-0 -v 3.0.1-2

This would definitely be useful, maybe even to a single test level. Let me work on adding handling for that. And yeah I'll flesh out some examples - I had the vague idea that each test would be its own python module, but didn't think about the sub-groupings of certificates. So good questions, I need to work with @ajcaldelas to get the certificate structure thought through & how we can populate everything.

DGonzalezVillal:
Should we close this?

PR #247, opened 2026-05-28
Include ID block when launching guests

markg-github:
This PR is motivated by the need for an SEV derived keys test in sev-certify/verify. SEV ID blocks include fields that affect what are valid inputs to SEV key derivation. If an ID block isn't used when an SEV guest is launched, valid key derivation inputs are limited.

An ID block includes expected measurement, guest policy, Family ID, Image ID, etc. The expected measurement must match the HW/FW-calculated measurement. Note that sev-certify has always calculated an expected measurement even though it's only now, with this PR, that an ID block is included when launching guests. Ignoring the ID block, sev-certify used the expected measurement as "host data". This is still the case.

ID blocks are signed, but they can be self-signed, which is what this PR does.

### DIFF
--- a/sev_verify/attestation_report.py
+++ b/sev_verify/attestation_report.py
@@ -1,0 +1,181 @@
+"""Parse the SEV-SNP ATTESTATION_REPORT binary structure.
+
+Reading ``report.bin`` directly, rather than regexing ``snpguest display
+report``, removes a dependency on a CLI's human-readable output format. The
+binary layout is fixed by the SEV-SNP ABI and, unlike the CLI text, the report
+is *self-describing*: VERSION is the first four bytes, so every report states
+which layout it uses and can be checked on the spot.
+
+Layout is version-dependent in practice — the report version tracks firmware,
+which tracks CPU generation — but versions have only ever *appended* fields.
+Everything below 0x188 is common to v2 and v3; the CPUID family/model/stepping
+triple at 0x188 exists only in v3+.
+
+Every offset here was validated against a real v3 report from an EPYC 9654
+(Genoa, CPUID 19h/11h), cross-checked against independently known values:
+GUEST_SVN/POLICY/FAMILY_ID/IMAGE_ID against the values the ID block was built
+with, REPORTED_TCB against ``snphost ok``, AUTHOR_KEY_DIGEST against the known
+all-zero author key, and CPUID against the CPU model.
+"""
+
+from __future__ import annotations
+
+import struct
+from dataclasses import dataclass
+from pathlib import Path
+
+#: ATTESTATION_REPORT is a fixed-size structure.
+REPORT_SIZE = 1184
+
+#: Report versions whose layout we read. v3 is verified on hardware; v2 shares
+#: the same layout for every field below 0x188.
+KNOWN_VERSIONS = frozenset({2, 3})
+
+# Field offsets. See module docstring for how these were validated.
+_OFF_VERSION = 0x000
+_OFF_GUEST_SVN = 0x004
+_OFF_POLICY = 0x008
+_OFF_FAMILY_ID = 0x010
+_OFF_IMAGE_ID = 0x020
+_OFF_VMPL = 0x030
+_OFF_REPORT_DATA = 0x050
+_OFF_MEASUREMENT = 0x090
+_OFF_HOST_DATA = 0x0C0
+_OFF_ID_KEY_DIGEST = 0x0E0
+_OFF_AUTHOR_KEY_DIGEST = 0x110
+_OFF_REPORT_ID = 0x140
+_OFF_REPORTED_TCB = 0x180
+_OFF_CPUID_FAM = 0x188  # v3+
+
+_LEN_ID = 16
+_LEN_MEASUREMENT = 48
+_LEN_DIGEST = 48
+_LEN_REPORT_DATA = 64
+_LEN_HOST_DATA = 32
+_LEN_REPORT_ID = 32
+
+
+class ReportError(Exception):
+    """Base class for attestation report problems."""
+
+
+class ReportMalformed(ReportError):
+    """The file is not a well-formed ATTESTATION_REPORT."""
+
+
+class ReportUnsupportedVersion(ReportError):
+    """The report declares a version whose layout we have not validated."""
+
+
+@dataclass(frozen=True)
+class TcbVersion:
+    """Decoded SNP TCB_VERSION — the same four values ``snphost ok`` prints."""
+
+    bootloader: int
+    tee: int
+    snp: int
+    microcode: int
+
+    @classmethod
+    def from_bytes(cls, raw: bytes) -> TcbVersion:
+        # byte 0 BOOT_LOADER, byte 1 TEE, bytes 2-5 reserved,
+        # byte 6 SNP, byte 7 MICROCODE.
+        return cls(bootloader=raw[0], tee=raw[1], snp=raw[6], microcode=raw[7])
+
+    def __str__(self) -> str:
+        return (
+            f"bootloader={self.bootloader} tee={self.tee} "
+            f"snp={self.snp} microcode={self.microcode}"
+        )
+
+
+@dataclass(frozen=True)
+class AttestationReport:
+    """The fields of an ATTESTATION_REPORT that we read."""
+
+    version: int
+    guest_svn: int
+    policy: int
+    family_id: bytes
+    image_id: bytes
+    vmpl: int
+    report_data: bytes
+    measurement: bytes
+    host_data: bytes
+    id_key_digest: bytes
+    author_key_digest: bytes
+    report_id: bytes
+    reported_tcb: TcbVersion
+    #: (family, model, stepping) — v3+ only, None on older reports.
+    cpuid: tuple[int, int, int] | None
+
+    @property
+    def id_block_used(self) -> bool:
+        """True when ID_KEY_DIGEST is set, i.e. the guest launched with an ID block."""
+        return any(self.id_key_digest)
+
+
+def parse(data: bytes) -> AttestationReport:
+    """Parse raw report bytes.
+
+    Raises:
+        ReportMalformed: wrong size.
+        ReportUnsupportedVersion: layout not validated for that version.
+    """
+    if len(data) != REPORT_SIZE:
+        raise ReportMalformed(
+            f"expected a {REPORT_SIZE}-byte ATTESTATION_REPORT, got {len(data)} bytes"
+        )
+
+    (version,) = struct.unpack_from("<I", data, _OFF_VERSION)
+    if version not in KNOWN_VERSIONS:
+        raise ReportUnsupportedVersion(
+            f"report declares VERSION {version}; this parser has been validated "
+            f"for {sorted(KNOWN_VERSIONS)}"
+        )
+
+    (guest_svn,) = struct.unpack_from("<I", data, _OFF_GUEST_SVN)
+    (policy,) = struct.unpack_from("<Q", data, _OFF_POLICY)
+    (vmpl,) = struct.unpack_from("<I", data, _OFF_VMPL)
+
+    def field(off: int, length: int) -> bytes:
+        return data[off:off + length]
+
+    cpuid = None
+    if version >= 3:
+        cpuid = (
+            data[_OFF_CPUID_FAM],
+            data[_OFF_CPUID_FAM + 1],
+            data[_OFF_CPUID_FAM + 2],
+        )
+
+    return AttestationReport(
+        version=version,
+        guest_svn=guest_svn,
+        policy=policy,
+        family_id=field(_OFF_FAMILY_ID, _LEN_ID),
+        image_id=field(_OFF_IMAGE_ID, _LEN_ID),
+        vmpl=vmpl,
+        report_data=field(_OFF_REPORT_DATA, _LEN_REPORT_DATA),
+        measurement=field(_OFF_MEASUREMENT, _LEN_MEASUREMENT),
+        host_data=field(_OFF_HOST_DATA, _LEN_HOST_DATA),
+        id_key_digest=field(_OFF_ID_KEY_DIGEST, _LEN_DIGEST),
+        author_key_digest=field(_OFF_AUTHOR_KEY_DIGEST, _LEN_DIGEST),
+        report_id=field(_OFF_REPORT_ID, _LEN_REPORT_ID),
+        reported_tcb=TcbVersion.from_bytes(field(_OFF_REPORTED_TCB, 8)),
+        cpuid=cpuid,
+    )
+
+
+def read(path: Path) -> AttestationReport:
+    """Read and parse an ATTESTATION_REPORT file.
+
+    Raises:
+        ReportMalformed: file missing or wrong size.
+        ReportUnsupportedVersion: layout not validated for that version.
+    """
+    try:
+        data = path.read_bytes()
+    except FileNotFoundError as exc:
+        raise ReportMalformed(f"{path.name} not found") from exc
+    return parse(data)
--- a/sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
+++ b/sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
@@ -20,6 +20,7 @@ Pulled files and analysis output for this test go under ``ctx.artifact_dir``
 import subprocess
 from pathlib import Path
 
+from sev_verify.cvm_props import MeasurementError, read_measurement
 from sev_verify.models import BaseStep, Step, StepContext, StepHandlerResult
 from sev_verify.vm_profile import VMProfile, VMProfileError
 
@@ -82,10 +83,13 @@ def verify_report_fields(ctx: StepContext) -> StepHandlerResult:
     to values computed in earlier ``callable`` or ``host`` steps.
     """
     report_file = ctx.artifact_dir / "report.bin"
-    measurement_file = ctx.artifact_dir / "guest_measurement.txt"
     request_file = ctx.artifact_dir / "request.bin"
 
-    expected_measurement = measurement_file.read_text().strip()
+    try:
+        expected_measurement = f"0x{read_measurement(ctx.artifact_dir)}"
+    except MeasurementError as exc:
+        return StepHandlerResult(exit_code=1, stderr=str(exc))
+
     request_data = "0x" + str(request_file.read_bytes().hex())
     result = subprocess.run(
         [
--- a/sev_verify/cert_tests/c3_0/c3_0_0_2/id_block_test.py
+++ b/sev_verify/cert_tests/c3_0/c3_0_0_2/id_block_test.py
@@ -1,0 +1,362 @@
+"""id_block_test: Verify ID block acceptance, report field binding, and rejection.
+
+Positive path: launch an SEV-SNP guest with a valid ID block, fetch the
+attestation report, and verify that the hardware report reflects the ID block
+fields (guest_svn, policy, family_id, image_id).
+
+Negative path: attempt three launches that must fail:
+  1. ID block with a corrupted measurement (digest mismatch)
+  2. Policy incompatible with the platform (SMT=0 on an SMT-active host)
+  3. Impossibly high ABI major version (ABI_MAJOR=255)
+"""
+
+from __future__ import annotations
+
+import os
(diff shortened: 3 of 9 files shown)


modules/launch/host/generate-id-block/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+1. Reads the guest measurement from `guest_measurement.txt` (48-byte SHA-384,
+   written as `0x<96 hex chars>`)
+2. Decodes the ID block template from `id-block.b64` (must be exactly 96 bytes)
+3. Patches bytes 0–47 (the `ld` field) with the actual guest measurement
+4. Generates an ephemeral P-384 key pair
+5. Signs the patched ID block with ECDSA-P384-SHA384
+6. Writes the signed ID block back to `id-block.b64`
+7. Writes a 4096-byte `ID_AUTH_INFO` structure to `id-auth.b64` containing the
+   signature and the ephemeral public key (no author key)
```
DGonzalezVillal: This is a good approach. I don't mind keeping this approach but snpguest actually provides functionality to generate the id-block and the id-auth. Also allowing to pass the measurement, familiy id, svn and policy you want to load into the id-block.

tools/id-block-README.md
```diff
@@ -0,0 +1,74 @@
+# SEV-SNP ID Block Builder (`tools/id-block-builder.html`)
+
+Browser-based tool for building unsigned ID block templates for SEV-SNP guest
+launch. Opens locally — no server needed.
```
DGonzalezVillal: That's a really cool tool! Again I'm sorry I didn't let you know that snpguest could actually calculate it, and that will probably work best for testing, but I wonder if there's somewhere we can place this work.

sev_verify/cert_tests/common/snp_ok.py
```diff
@@ -47,7 +47,7 @@ def steps() -> list[BaseStep]:
     return [
         Step.for_host(
             name="snphost ok",
-            type="required",
+            type="info",
```
DGonzalezVillal: Is there a particular reason you changed this??

DGonzalezVillal (commented):
Just if possible change the type for snphost ok back to what it was and then it looks good to me.

sev_verify/cvm_props.py
DGonzalezVillal, line 1: Maybe we rename this to reference_values.py and we can put functions here that are relevant to that. id_block, measurement, etc.

markg-github: Good idea, though I don't know that most people directly associate an ID block with reference values, which I realize was only a suggestion.

sev_verify/cert_tests/common/snp_ok.py
markg-github: I have an old TCB and if I don't change this, testing doesn't get very far. I'll change back, and update my system.

sev_verify/cvm_props.py
markg-github: The new name is cvm_props.py.

sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
```diff
@@ -82,10 +83,13 @@ def verify_report_fields(ctx: StepContext) -> StepHandlerResult:
     to values computed in earlier ``callable`` or ``host`` steps.
     """
     report_file = ctx.artifact_dir / "report.bin"
-    measurement_file = ctx.artifact_dir / "guest_measurement.txt"
     request_file = ctx.artifact_dir / "request.bin"
 
-    expected_measurement = measurement_file.read_text().strip()
+    try:
+        expected_measurement = f"0x{read_measurement(ctx.artifact_dir)}"
```
amd-aliem, line 89: Just a note that it might be good future-proofing to having the path not hardcoded - doing so results in exactly one measurement file per test. If there are multiple VM launches it'll get overwritten (harder to debug), and if we have concurrent VMs it would also prevent having multiple measurements. But maybe we just open an issue since it was already hardcoding this before this PR.

sev_verify/cert_tests/c3_0/c3_0_0_2/id_block_test.py
```diff
@@ -0,0 +1,362 @@
...
+            timeout=300,
+        ),
+        Step.for_vm_stop(
+            name="Stop VM (after bad measurement)",
+            type="info",
+            timeout=60,
+        ),
+
+        # ── Negative: incompatible policy (SMT=0 on SMT-active host) ──
+        Step.for_callable(
```
amd-aliem, line 326: If I'm reading it right it looks like set_incompatible_policy errors if the host has SMT disabled. I would prefer not to add 'SMT must be enabled' as a prereq to this test suite.

sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
```diff
@@ -20,6 +20,7 @@
 import subprocess
 from pathlib import Path
 
+from sev_verify.cvm_props import MeasurementError, read_measurement
```
amd-aliem, line 23: import calculate_measurement too? you're using the local one still, below, in step 1 of this test. I think remove the calculate_measurement in this file.

sev_verify/cert_tests/c3_0/c3_0_0_2/id_block_test.py
```diff
@@ -0,0 +1,362 @@
...
+            name="Set bad measurement in ID block",
+            type="required",
+            handler="set_bad_measurement",
+            timeout=30,
+        ),
+        Step.for_vm_launch(
+            name="Launch with bad measurement (expect rejection)",
+            type="required",
+            expected_result="exit_code:1",
```
amd-aliem, line 316: it might be more accurate to look for the reason for failure instead of accepting any VM launch error (could be something other than ID block). But I don't know how difficult that would be to implement, so I'll leave it up to you. If we didn't change VM profile settings from the previous tests it's unlikely that something else would case a failure.

sev_verify/cvm_props.py
```diff
@@ -0,0 +1,229 @@
...
+
+def generate_id_block(ctx: StepContext) -> StepHandlerResult:
+    """Generate an ID block and auth block for the current guest measurement.
+
+    Reads guest_measurement.txt from ctx.artifact_dir (written by
+    calculate_measurement). Generates two ephemeral P-384 key pairs, invokes
+    snpguest generate id-block, and updates ctx.profile with the resulting
+    id_block and id_auth values so that vm_launch passes them to QEMU.
+
+    ID block metadata is read from environment variables with the same defaults
```
amd-aliem, line 149: What is this `generate-id-block` service? Is this comment still relevant?

sev_verify/attestation_report.py
```diff
@@ -0,0 +1,181 @@
...
+    bootloader: int
+    tee: int
+    snp: int
+    microcode: int
+
+    @classmethod
+    def from_bytes(cls, raw: bytes) -> TcbVersion:
+        # byte 0 BOOT_LOADER, byte 1 TEE, bytes 2-5 reserved,
+        # byte 6 SNP, byte 7 MICROCODE.
+        return cls(bootloader=raw[0], tee=raw[1], snp=raw[6], microcode=raw[7])
```
amd-aliem, line 83: no fmc/turin support?

sev_verify/attestation_report.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+@dataclass(frozen=True)
+class TcbVersion:
+    """Decoded SNP TCB_VERSION — the same four values ``snphost ok`` prints."""
+
+    bootloader: int
+    tee: int
+    snp: int
+    microcode: int
```
amd-aliem, line 77: no fmc?

sev_verify/attestation_report.py
```diff
@@ -0,0 +1,181 @@
+"""Parse the SEV-SNP ATTESTATION_REPORT binary structure.
```
amd-aliem, line 1: I don't like the idea of re-implementing report parsing and re-defining all of the report fields. I think if we need better interfaces to the report, we should update snpguest. We can add a `--json` maybe. We're missing out on exercising our own `sev` crate and user-facing tool, and we're increasing our own work whenever a new report version comes out (have to update it here as well).

We can perhaps continue with this approach in the short term if there's no way around it, but I really do not want this to be a long-term approach.

cc @DGonzalezVillal

sev_verify/cert_tests/c3_0/manifest.toml
```diff
@@ -19,3 +19,12 @@ description = "Test SNP_CONFIG and SNP_COMMIT host commands"
...
 level = "3.0.0-1"
+
+# ── Level 3.0.0-2 ───────────────────────────────────────────
+
+[[tests]]
+name = "id-block-test"
+description = "Verify ID block acceptance, report field binding, and launch rejection"
+module = "cert_tests.c3_0.c3_0_0_2.id_block_test"
+scope = "mixed"
+level = "3.0.0-2"
```
amd-aliem, line 30: 3.0.0-1, I think we're still operating under the 'group tests under one dash release' methodology

amd-aliem (changes requested):
A few comments - also could you please show a working e2e output (link an issue in your fork where you ran this new code).

markg-github:
Thank you very much for the comments, @amd-aliem. Please forgive any short replies like ACK, they just mean understood and update on the way.

sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
markg-github: ACK

sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
markg-github: I'll open an issue.

sev_verify/cert_tests/c3_0/c3_0_0_2/id_block_test.py
markg-github: ACK

My small PR (#289) that merged last week facilitates the improvement alluded to in the comment.

sev_verify/cert_tests/c3_0/c3_0_0_2/id_block_test.py
markg-github: Yeah, I'll look again at the other *allowed bits in the guest policy. It would be nice if, for a given, less permissive guest policy, some QEMU command lines worked (launched) and some didn't vs. being at the mercy of a platform setting like SMT. If none of the bits work this way, do you think it would be better to skip this sub-test when SNP is disabled or not have the sub-test at all? I agree that a change is required either way.

sev_verify/cert_tests/c3_0/manifest.toml
markg-github: ACK

sev_verify/attestation_report.py
markg-github: I'll open an issue (snpguest) or two (and sev) in virtee.

sev_verify/attestation_report.py
markg-github: ACK

sev_verify/attestation_report.py
markg-github: ACK

sev_verify/cvm_props.py
markg-github: ACK - stale comment

sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
markg-github: On second thought, I think I'll add a comment to #305 and possibly to #268.

sev_verify/attestation_report.py
markg-github: @amd-aliem , you opened #304 and I added a comment related to key/field names to it. Should I still open an issue or two in virtee?

sev_verify/attestation_report.py
markg-github: 37e5602

I did decide, as Claude would say, to fail loudly on CPUs or report versions that I didn't test. What do you think?

markg-github:
> A few comments - also could you please show a working e2e output (link an issue in your fork where you ran this new code).

Here's one: https://github.com/markg-github/sev-certify/issues/209

markg-github:
> Just if possible change the type for snphost ok back to what it was and then it looks good to me.

Done

markg-github:
A few more changes. Unfortunately, one has a relatively high line count: the patch wasn't consistent with how it managed some of the values added to the ID block. Passing cert with all the changes: https://github.com/markg-github/sev-certify/issues/214

images/host-centos-10/mkosi.conf
```diff
@@ -22,5 +22,6 @@ Packages=
 	xxd
 	python3
 	python3-pip
```
amd-aliem: I noticed that Fedora-44 isn't included here, but still appears to run fine (based on your [example issue](https://github.com/markg-github/sev-certify/issues/225)).

I won't ask you to do this here, but imo we should open an issue to make use of the common host packages file where possible to not have to re-define them a bunch of times & potentially miss one.

sev_verify/attestation_report.py
amd-aliem: Yes, I think this is fine for now - I do see your note that this is a short-term solution until snpguest has a machine-readable output, which is good. I think I remember @DGonzalezVillal mentioning that rust supports bindings to other languages too - if we do that for the `sev` package directly to python that may be another option.

sev_verify/attestation_report.py
amd-aliem: Thanks, let's chat with the team about which we prefer (snpguest or a sev rust->python binding). I've added it to the agenda for next meeting.

sev_verify/cert_tests/c3_0/manifest.toml
```diff
@@ -20,3 +20,10 @@ module = "cert_tests.c3_0.c3_0_0_1.snphost_config_commit"
 scope = "mixed"
 level = "3.0.0-1"
 host_changes = true
+
+[[tests]]
+name = "id-block-test"
+description = "Verify ID block acceptance, report field binding, and launch rejection"
+module = "cert_tests.c3_0.c3_0_0_2.id_block_test"
```
amd-aliem: please update to `3_0_0_1` here, and the path for `id_block_test.py`

sev_verify/cvm_props.py
```diff
@@ -0,0 +1,319 @@
...
+from cryptography.hazmat.primitives.asymmetric import ec
+from cryptography.hazmat.primitives.serialization import (
+    Encoding,
+    NoEncryption,
+    PrivateFormat,
+)
+
+from .models import StepContext, StepHandlerResult
+# from .vm_profile import VMProfile, VMProfileError
+from .vm_profile import VMProfileError
```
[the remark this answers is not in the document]
amd-aliem: @markg-github I agree, please remove the comment

sev_verify/cert_tests/c3_0/c3_0_0_2/id_block_test.py
```diff
@@ -0,0 +1,475 @@
...
+
+    # ── Negative: incompatible policy (SMT=0 on SMT-active host) ──
+    #
+    # Clearing the SMT bit only produces a rejection on a host where SMT is
+    # actually active, so elsewhere there is nothing to assert.  The case is
+    # left out of the step list rather than run and failed, with an info step
+    # in its place: a case that vanishes silently is indistinguishable from one
+    # that passed.  Reporting it as "pass" does overload that outcome — a
+    # first-class per-step "not applicable on this platform" result would say
+    # so plainly, and is the better home for this once one exists.
```
amd-aliem: @markg-github please create an issue for this enhancement (new step type).

Also please include a step-type to handle an expected VM-is-off, and fail and shut down the VM if not. This is to handle seeing `ERR!` in a correctly functioning test ([from your issue](https://github.com/markg-github/sev-certify/issues/214)):
```
   ├─ Launch with SMT-incompatible policy (expect rejection) ······· 2.0s [PASS]
   ├─ Stop VM (after SMT policy) ···································· 0ms [ERR!]
   │     vm_stop: no running guest (run vm_launch or a guest step first)
   ├─ Set impossible ABI version ··································· 12ms [PASS]
   ├─ Launch with impossible ABI version (expect rejection) ········ 2.0s [PASS]
   └─ Stop VM (after ABI version) ··································· 0ms [ERR!]
```
This is fine for now since we want to get this test merged, but should be cleaned up to show clean "PASS" steps in the future.

amd-aliem (changes requested):
Looks almost good to me - will approve once you move the `c3_0_0_2/id_block_test.py` -> `c3_0_0_1`. Lots of future work I think we should do, but nothing else that would block merging this, imo. 
Thanks for answering all my review comments!

markg-github:
Address @amd-aliem 's most recent comments.

Also, add support for more CPU models.

sev_verify/attestation_report.py
```diff
@@ -0,0 +1,530 @@
...
+===========  =====================  ====================
+byte         Milan / Genoa          Turin / Venice
+===========  =====================  ====================
+0            BOOT_LOADER            FMC
+1            TEE                    BOOT_LOADER
+2            --                     TEE
+3            --                     SNP
+6            SNP                    --
+7            MICROCODE              MICROCODE
+===========  =====================  ====================
```
DGonzalezVillal: FYI Venice is out and it's different:

0 FMC
1 TEE
2 SNP

sev_verify/attestation_report.py
```diff
@@ -0,0 +1,530 @@
...
+all-zero author key, and CPUID against the CPU model.
+
+The second used a **version 5** report from an EPYC 9575F (Turin, CPUID
+1Ah/02h), which validated the Turin TCB layout for the first time. Its
+REPORTED_TCB bytes were ``0103020600000062``, decoding under the Turin layout to
+``bootloader=3 tee=2 snp=6 microcode=98 fmc=1`` — matching ``snphost show tcb``
+exactly. Decoded under the legacy layout the same bytes give
+``bootloader=1 tee=3 snp=0 microcode=98`` with no FMC: plausible values, silently
+wrong, which is precisely the failure this generation gate exists to prevent.
+That report also confirmed v5 moved none of the fields read here.
```
DGonzalezVillal: Helpful to know you tested, but unnecessary for documentation. It's just making this too verbose.

DGonzalezVillal:
```suggestion
```

sev_verify/attestation_report.py
DGonzalezVillal: This works for now

sev_verify/cvm_props.py
```diff
@@ -0,0 +1,327 @@
...
+
+_MEASUREMENT_FILE = "guest_measurement.txt"
+_ID_BLOCK_FILE = "id-block.b64"
+_ID_AUTH_FILE = "id-auth.b64"
+
+DEFAULT_FAMILY_ID = "sev-certify-fam0"
+DEFAULT_IMAGE_ID = "sev-certify-img0"
+DEFAULT_GUEST_SVN = "48"
+DEFAULT_POLICY = "0xb0000"
```
DGonzalezVillal: Default policy in SNP is 0x30000

DGonzalezVillal (approved):
Just one comment on the DEFAULT POLICY you have. I also made a comment on Venice, But you don't have to add it here.

sev_verify/cvm_props.py
markg-github: Default in what sense? Default here is as opposed to override (eg, env var) and is intended to be most permissive, regardless of any other defaults.

That said, given how SEV "debug mode" works, which I didn't understand at the time, 0x30000 and 0xB0000 are equally permissive.

markg-github:
Based on comments and since I'd been leaning in the same direction, I'm going to go through the PR again with a goal to make code and git comments more human friendly.

amd-aliem (approved):
Please don't forget to open the issues!

amd-aliem (approved):
lgtm, thanks!

images/host-centos-10/mkosi.conf
markg-github: I may have rebased after Fedora-44 was added and then failed to include Fedora-44. 

To your main point, the "common host packages file" already exists, right? I think AI and I were anticipating different names for the package across our distros.

PR #252, opened 2026-06-17
Add changes to sev_verify to allow for vm launches and complete testing

DGonzalezVillal:
# Changes Made to Amanda's Original PR

## Step Type vs. Step Kind

`step type` tells us the severity of the step: `setup`, `required`, or `info`.

Depending on the step type, execution will either continue or stop after a failure.

`step kind` tells us how the step will be executed:

- `vm_launch`: Launches the `VMProfile` defined in the test.
- `vm_stop`: Stops the running guest.
- `guest`: Sends a command to be run in the guest's shell through vsock.
- `guest_pull`: Pulls files from the guest into the host.
- `host`: Runs a command in the host's shell.
- `callable`: Runs a defined function as a step.

## VM Profile

Defines the virtual machine settings for the VM that will be used for testing.

The `--qemu-binary` / `--qemu` and `--ovmf` flags override the default QEMU and OVMF binaries when set.

## QEMU and OVMF Flags

Added flags to pass custom QEMU or OVMF binaries in case the user wants to use their own instead of the default OS-provided ones.

## Artifacts Flag

Added a flag to pass the directory location where artifacts used during testing will be stored.

The flag is:

```bash
--artifacts-dir
```
The default value is

```bash
./artifacts
````

The artifact directory layout is:
```bash
<dir>/<cert version>/<level>/<test_name>/
```

Prequisite artifacts use:
```bash
<dir>/prereqs/<test_name>/
```

Host steps receive the artifact directory through:
```bash
$SEV_VERIFY_ARTIFACT_DIR
```

Callable steps use:
```bash
ctx.artifact_dir
```
For guest_pull steps, a relative host_dest path is stored under the per-test artifact directory.

## End-to-End Attestation Test

The attestation-test shows what a full end-to-end test will look like.


images/guest-centos-10/mkosi.conf
```diff
@@ -18,3 +18,4 @@ Packages=
 	jq
 	dnf
 	xxd
+	python3
```
amd-aliem, line 21: Don't think it's much of a problem after comparing sizes, but just a note that this python3 + vsock change is about +12M (227M->239M). If image size becomes a problem in the future could maybe look into a compiled language for vsock_exec.

sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
```diff
@@ -0,0 +1,210 @@
...
+from sev_verify.models import BaseStep, Step, StepContext, StepHandlerResult
+from sev_verify.vm_profile import VMProfile, DEFAULT_OVMF_CANDIDATES
+
+vm_profile = VMProfile(
+    image_path="",
+    memory_mb=2048,
+)
+
+def calculate_measurement(ctx: StepContext) -> StepHandlerResult:
+    """
```
amd-aliem, line 32: This can be simplified to use `ctx.profile.resolved_ovmf_path()` since you added the ovmf step to the "Verify SNP" test

sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
```diff
@@ -0,0 +1,210 @@
...
+            host_dest="request.bin",
+            timeout=120,
+        ),
+        Step.for_host(
+            name="Fetch certificate chain from kds",
+            type="setup",
+            command='snpguest fetch ca pem "$SEV_VERIFY_ARTIFACT_DIR" -r "$SEV_VERIFY_ARTIFACT_DIR/report.bin"',
+            timeout=60,
+        ),
+        Step.for_host(
```
amd-aliem, line 168: I kind of get a lot of 429s when running - in the future we may want to create a folder in `artifacts` for certificate caching. It's not a big deal right now b/c we're only fetching during this particular test, but in the future if other tests also want to check attestation reports it may become a real problem.

sev_verify/vm_profile.py
```diff
@@ -0,0 +1,392 @@
...
+        normalized: dict[str, Any] = {}
+        for key, value in data.items():
+            field_name = aliases.get(key, key.replace("-", "_"))
+            normalized[field_name] = value
+        known = {f.name for f in fields(cls)}
+        return cls(**{k: v for k, v in normalized.items() if k in known})
+
+    def resolved_ovmf_path(self) -> str:
+        """"
+        Verify provided OVMF path is presnet.
```
amd-aliem: `presnet` -> `present`

sev_verify/vm_profile.py
```diff
@@ -0,0 +1,392 @@
...
+    if len(decoded) != HOST_DATA_SIZE:
+        raise VMProfileError(
+            f"host_data must be exactly {HOST_DATA_SIZE} bytes; got {len(decoded)} bytes"
+        )
+
+
+@dataclass(frozen=True)
+class VMProfile:
+    """Launch-time configuration for an SEV-SNP guest."""
+    # QEMU varfiables with non-default values
```
amd-aliem: `varfiables` -> `variables`

amd-aliem:
Only small comments, I think the VM impl is great, & the interface to add a test feels good imo. after this is merged, some next steps for us:
- add document about how to add a test (including updating certificate matrix table with feature/test overview)
- add sev_verify to host image
- modify beacon to read new result files
- maybe expand the md result file to have more information, like system info?
- maybe add vcek cert cache to avoid kds rate limiting? might be needed if future tests need to also verify attestation sigs (host config/commit probably will for the guest side tests).

modules/build/guest/vsock/mkosi.extra/usr/local/lib/systemd/system/cert-vsock-agent.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Certification test guest vsock command agent
+After=network.target
```
markg-github: Claude is saying this is unnecessary since vsock has no network dependency.
```suggestion
```

images/guest-centos-10/mkosi.conf
DGonzalezVillal: That's true. Since we are also doing a lot less work on the guest side, in reality we can remove some of the packages out like jq, dnf, xxd, journal-remote. A lot of that stuff was there for the previous architecture.


The guest images could probably ship with kernel python and snpguest and should be good to go. And a network for kds outreach.

DGonzalezVillal: I had not strip them, but I'll reduce them to what I think is the minimum requirements

sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
DGonzalezVillal: I don't know if it highlighted the wrong lines, but I'm assuming you're saying we can use the resolved_ovmf_path to get the OVMF instead of using the logic there present. I can make that change

sev_verify/README.md
```diff
@@ -0,0 +1,80 @@
+# sev_verify
+
+Host-side testing harness for SEV-SNP certification. Reads TOML manifests that declare which tests to run, imports per-test Python modules that define executable steps, and orchestrates execution across host and guest environments.
+
```
markg-github, line 5:
```suggestion
sev-verify uses a non-secure vsock channel between the host and the guest, which is launched as a CVM. Given the purpose of sev-certify, this is acceptable. The vsock channel properties are properties of the guest and the guest is purpose-built for sev-certify. As built, there is no incentive for an attacker to take advantage of the security weakness of the vsock channel.
```

sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
DGonzalezVillal: Yeah I get those all the time too. I wonder if the code might have a bug somewhere where it does several consecutive requests to the kds.

markg-github (commented):
I used the suggestion feature of PR comments, but didn't pick the Request changes option (they're independent) since the suggestions are minor.

images/guest-centos-10/mkosi.conf
DGonzalezVillal: - strip image
- debug image

sev_verify/cert_tests/c3_0/c3_0_0_0/attestation_test.py
DGonzalezVillal: - If it becomes too much of an issue in automated testing. Add a 429 handler
-

modules/build/guest/vsock/mkosi.extra/usr/local/lib/systemd/system/cert-vsock-agent.service
DGonzalezVillal: You are correct. I only put that there to verify that this gets executed post-boot, but it seems that by only having 

[Install] 
WantedBy=multi-user.target

should be enough to ensure that.

sev_verify/README.md
DGonzalezVillal: added

amd-aliem (approved):

PR #255, opened 2026-06-29
fix: add vsock kmods back to ubuntu 26.04

amd-aliem:
Ubuntu 26.04 guest images were timing out when booting with the sev_verify harness. The cause was traced to a missing vsock kmod - add it back. Other images not affected because they don't use the `KernelModulesExclude=.*` setting

### DIFF
--- a/images/guest-ubuntu-26.04/mkosi.conf
+++ b/images/guest-ubuntu-26.04/mkosi.conf
@@ -15,7 +15,6 @@ Packages=
 	systemd
 	systemd-boot-efi
 	systemd-resolved
-	network-manager
 	plymouth
 	login.defs
 	login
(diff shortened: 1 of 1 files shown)


DGonzalezVillal (approved):

markg-github:
**Automated review notes** (post-merge, for the record)

The vsock fix itself looks correct — `net/vmw_vsock/` is the right granularity for the directory glob and accurately captures the full virtio vsock transport stack. The root cause explanation (26.04 uniquely uses `KernelModulesExclude=.*`, so vsock must be explicitly whitelisted) is well-documented.

**One flag worth noting:** this PR also removes `network-manager` from the `Packages` list, which isn't mentioned in the description. The sibling Ubuntu guest configs (25.04, 25.10) don't include `network-manager` either, so this likely brings 26.04 into alignment — but if anything in the 26.04 boot path expects NetworkManager to manage interfaces, it would silently fail. Worth confirming that was intentional and tested.

**Open question:** the description says "add it back" — when was vsock previously removed? If it was an over-aggressive `KernelModulesExclude=.*` cleanup, other modules may be missing from the 26.04 image that haven't surfaced as failures yet.

PR #260, opened 2026-07-08
build: add tags local build option

amd-aliem:
Building locally often gets you rate-limited by GH, since we do not support authenticated calls. Add a `make tags` option that pre-fetches the tags. Due to how mkosi handles environments, this is done by creating an environment file and passing it to the mkosi command.

This commit does not affect the normal CI image builds, as the images/Makefile is just for local development.

### DIFF
--- a/images/Makefile
+++ b/images/Makefile
@@ -1,7 +1,30 @@
 IMAGES := $(sort $(notdir $(wildcard host-* guest-*)))
 CLEAN_IMAGES := $(addprefix clean-,$(IMAGES))
 
-.PHONY: all clean list status $(IMAGES) $(CLEAN_IMAGES)
+GH_AUTH := $(if $(GITHUB_TOKEN),-H "Authorization: Bearer $(GITHUB_TOKEN)")
+GH_LATEST = curl -sf $(GH_AUTH) https://api.github.com/repos/$(1)/releases/latest | jq -r '.tag_name'
+
+.PHONY: all clean list status tags $(IMAGES) $(CLEAN_IMAGES)
+
+tags:
+	@set -e; \
+	trap 'rm -f .tags.tmp' EXIT; \
+	fetch() { \
+		tag=$$($(call GH_LATEST,$$1)); \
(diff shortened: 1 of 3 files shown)


images/Makefile
```diff
@@ -19,7 +31,7 @@ status:
 	done
 
 $(IMAGES):
-	mkosi --image-id=$@ -C $@ build
+	mkosi --image-id=$@ -C $@ --env-file=../.tags build
```
markg-github, line 45: **F1. [MEDIUM] Error handling — `--env-file=../.tags` is unconditional but Make doesn't enforce the prerequisite**

The updated README correctly documents `make tags` as step 1 before building, so the intended workflow is clear. But the Makefile itself doesn't express this: there's no `$(IMAGES): | .tags` rule, so Make won't catch a missing `.tags` with a helpful error.

Previously `make host-fedora-41` worked standalone because the build scripts have `${VAR:-$(curl ...)}` fallbacks. Now it will fail with a cryptic mkosi error about a missing env-file.

A couple of options to align the Makefile with the README:
- **Order-only prerequisite**: `$(IMAGES): | .tags` — Make will tell the user to run `make tags` first
- **Conditional env-file**: `$(if $(wildcard .tags),--env-file=../.tags)` — preserves the old standalone-build behavior as a fallback

Either way the question is: should the Makefile enforce what the README documents, or should standalone builds (with curl fallback) keep working?

markg-github (commented):
Automated review — one medium finding about the mandatory .tags prerequisite posted inline. README documentation is clear; question is whether the Makefile should enforce the same flow.

markg-github (dismissed):
LGTM

images/Makefile
amd-aliem: That's a good catch, I was thinking to keep them separate in case you want to re-generate tags (if snphost had a new release or something), but didn't test what happened if you DONT run 'make tags' first. I'll fix it to enforce it as a prereq.

DGonzalezVillal (approved):

PR #261, opened 2026-07-09
fix: README column match in cert matrix workflow

amd-aliem:
The README table headers use [EPYC XXXX][cert-X.Y] format, not 'X.Y Certification'. The column lookup searched for '3.0 Certification' which never matched, silently skipping the README update while certifications.md was updated correctly.

Tested on my fork.

### DIFF
--- a/.github/workflows/update-certification-matrix.yml
+++ b/.github/workflows/update-certification-matrix.yml
@@ -157,7 +157,7 @@ jobs:
                   
                   if (headerIdx !== -1) {
                     const headers = readmeLines[headerIdx].split('|').map(h => h.trim()).filter(h => h);
-                    const readmeColumnIndex = headers.findIndex(h => h.includes(`${certVersion} Certification`));
+                    const readmeColumnIndex = headers.findIndex(h => h.includes(`EPYC ${procSeries}`));
                     
                     if (readmeColumnIndex !== -1) {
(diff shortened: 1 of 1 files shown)


.github/workflows/update-certification-matrix.yml
```diff
@@ -157,7 +157,7 @@ jobs:
                   
                   if (headerIdx !== -1) {
                     const headers = readmeLines[headerIdx].split('|').map(h => h.trim()).filter(h => h);
-                    const readmeColumnIndex = headers.findIndex(h => h.includes(`${certVersion} Certification`));
+                    const readmeColumnIndex = headers.findIndex(h => h.includes(`EPYC ${procSeries}`));
```
markg-github, line 160: **Open question:** Given the old code always silently skipped the README update, has this workflow ever actually modified the README? If not, this fix will cause it to start writing to the README for the first time.

markg-github (commented):
Automated review — fix looks correct. One minor consistency note and an open question posted inline.

markg-github (approved):
LGTM

.github/workflows/update-certification-matrix.yml
amd-aliem: Yes, in the past it did update the README, but at some point we changed the header. We do want the behavior of updating it though, to get auto-updates of the main matrix.

PR #262, opened 2026-07-10
docs: add overview of mkosi and systemd use

markg-github:
A bit of a placholder for now (and, therefore, a draft), but intended to do what the title says.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,0 +1,68 @@
+# Shared architecture notes (sev-certify → sev-performance)
+
+This document captures **shared design/implementation context** that existed before `sev-performance` (perf) was forked from `AMDEPYC/sev-certify`.
+
+It’s intended to be:
+
+- **Short and stable** (concepts, not step-by-step recipes)
+- Useful to both repos
+- A landing page that we can eventually move upstream to `sev-certify` once it settles
+
+## Why a workflow exists at all
+
+Both projects need a repeatable way to produce **bootable host and guest OS images** that:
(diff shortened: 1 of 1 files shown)


amd-aliem:
Overall doc name and scope seems like a good idea to me - though I don't know about having a shared doc w/ sev-performance, I'd think it'd be better to not have references to sev-performance (as that opens a whole other discussion on what that project does, which is imo not necessary to know about).

In this arch doc I'd want to link in how sev_verify is run and the various stages (i.e. boot.target -> system.target -> test.target -> stop.target) as well.

Also useful to have the diagram about dispatch, beacon, and github relationships (data flow) - I think @DGonzalezVillal had one.

Similar points about missing docs in docs/modules, though I assume you'd add that later.

markg-github:
> Overall doc name and scope seems like a good idea to me - though I don't know about having a shared doc w/ sev-performance, I'd think it'd be better to not have references to sev-performance (as that opens a whole other discussion on what that project does, which is imo not necessary to know about).
> 
> In this arch doc I'd want to link in how sev_verify is run and the various stages (i.e. boot.target -> system.target -> test.target -> stop.target) as well.
> 
> Also useful to have the diagram about dispatch, beacon, and github relationships (data flow) - I think @DGonzalezVillal had one.
> 
> Similar points about missing docs in docs/modules, though I assume you'd add that later.

Agree. It may only ever be a placeholder/reminder that it would be good to have an "architecture" (overview) doc.

PR #263, opened 2026-07-13
feat: add ID block support to sev_verify test harness (with example test)

markg-github:
Draft version of the ID block PR that includes an example test demonstrating usage.

The `attestation-id-block` test at cert level 3.0.0-2 shows how a test module uses the ID block framework: `calculate_measurement` and `generate_id_block` callable steps (from `sev_verify.id_block`) prepend the standard attestation steps. The hardware-signed attestation report confirms family_id, image_id, svn, and policy are reflected correctly from the ID block.

The ready-for-review version of this PR (without the example test) is PR #247. This draft exists as a reference — the derived keys PR will be the first real consumer of the framework.


DGonzalezVillal (commented):
I like the idea of adding ID block tests, let's merge your other one first and then we can put this one in as well.

markg-github:
Changing my other ID block PR to include a better test, which makes this PR unnecessary.

PR #283, opened 2026-07-23
feat: convert snphost_config_commit to mixed-scope

amd-aliem:
Convert the host-only TCB config test into a mixed-scope test that verifies config changes are reflected in guest attestation reports, and adds a comprehensive doc about what SNP_COMMIT and SNP_CONFIG do + intended use.

Test by inspecting attestation reports in these cases:

1. Fresh VM (booted after config-set-lower): guest reported == host reported tcb guest current == host platform tcb signed by lowered host reported tcb
2. Live VM (same VM as 1 after config-reset): guest reported == restored host reported tcb guest current == host platform tcb

Only the alternate lower tcb VCEK is fetched, as normal is covered by 3.0.0-0 attestation test and don't want to trigger KDS rate-limit.

A final info-type config-reset ensures TCB cleanup regardless of test outcome.

Note:
- Implemented Copilot's suggestion to catch VMProfileError instead of VMLaunchError, as VMProfileError is the parent of VMLaunchError, and `profile.vm_launch()` can throw either of those. Prevents a crash if we run into the ProfileError.
- Example Certification Run: https://github.com/amd-aliem/sev-certify/issues/202


sev_verify/README.md
```diff
@@ -79,3 +79,9 @@ results/                 Output (gitignored)
 ## Requirements
 
 Python 3.11+ (uses `tomllib` from stdlib). No external packages.
+
+## ⚠️ Caution: tests that make permanent platform changes
+
+Most tests are transient — `snphost config set`/`reset`, VM launches, etc. reset on reboot. **One test makes an irreversible change:**
+
+- **`3.0.0-1` snphost-config-commit** runs `snphost commit`, which writes `CommittedTcb := CurrentTcb` into the PSP. This **permanently advances the platform's committed TCB floor**: it cannot be lowered (only raised by future firmware updates) and does **not** reset across reboots. After commit, the platform will refuse to boot firmware below the committed level. In normal operation the commit lands on the level already committed (a no-op), but on a machine where `CommittedTcb < CurrentTcb` it **will** advance the floor. The test records the committed floor before/after in its step output. See [`docs/features/tcb-config-commit.md`](../docs/features/tcb-config-commit.md) for details.
```
DGonzalezVillal: I actually think that TCB does reset on reboot, I need to double check on that. But I think on system reboots the fw will load the firmware to its original values again.

We should maybe also talk about platform changing tests? Maybe we don't want to include those?? Something to talk about.

DGonzalezVillal (commented):
Just one comment

amd-kkin (dismissed):
Thanks for adding the info!

sev_verify/README.md
amd-aliem: Confirmed w/ firmware team - you're right, it resets on reboot. Will update the docs & warnings 👍

sev_verify/cert_tests/c3_0/c3_0_0_1/snphost_config_commit.py
```diff
@@ -106,6 +154,206 @@ def verify_differ(_ctx: StepContext) -> StepHandlerResult:
...
+    current = sections["Current"]
+
+    dump = f"  Committed: {committed}\n  Current:   {current}"
+
+    if _tcb_fields_match(committed, current):
+        return StepHandlerResult(
+            exit_code=0,
+            stdout=f"Committed == Current - commit is a no-op on the floor.\n{dump}",
+        )
+
```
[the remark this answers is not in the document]
amd-aliem: fixed

modules/test/host/sev-verify/mkosi.extra/usr/local/lib/scripts/run-sev-verify.sh
```diff
@@ -6,9 +6,13 @@ LOG_FILE="${RESULTS_DIR}/sev-verify.log"
 
 mkdir -p "$RESULTS_DIR"
 
+# This is a dedicated, disposable test-host image (rebuilt/reprovisioned per
+# run), so boot-session-only config changes are fine — notably letting
+# `snphost commit` advance the committed TCB floor (it resets on reboot).
 python3 -m sev_verify \
 	/usr/local/lib/guest-image/guest.efi \
 	--output-dir "$RESULTS_DIR" \
+	--disposable-host \
```
DGonzalezVillal, line 15: How would you feel of instead of naming this as disposable host, we make the flag like --dangerous-tests or something similar. And under this flag we can run dangerous tests that may affect the permanent state of the machine. I guess that people may run this on non-disposable hosts, that's my only thought.

sev_verify/cert_tests/c3_0/c3_0_0_1/snphost_config_commit.py
```diff
@@ -147,62 +404,138 @@ def steps() -> list[BaseStep]:
...
-            name="config-reset-masks",
-            type="required",
-            command="snphost config reset",
+        # 14. Commit precondition (setup) — require Committed == Current so the
+        #     commit cannot advance the floor. A failure halts before the commit
+        #     below; `--disposable-host` downgrades it to a warning. See
+        #     verify_committed_equals_current.
+        Step.for_callable(
+            name="verify-committed-equals-current",
+            type="setup",
```
markg-github, line 514: type="setup" here is what makes the teardown skippable. On a setup failure the runner marks every remaining step skip and breaks out of the loop, so steps 15–17 never execute — including teardown-config-reset. The PR description's "a final info-type config-reset ensures TCB cleanup regardless of test outcome" doesn't hold on this path. (required failures are fine — those don't halt, so trailing info steps still run. Only setup does this.)

In practice the window is small: the expected failure here is provisional firmware without --disposable-host, which happens after step 9's config-reset already succeeded, so TCB is clean anyway. Actually stranding a lowered ReportedTcb needs step 9 to fail and report_after_reset.bin to be missing. The skipped steps also show up as skip in the results, so it isn't silent.

Still worth tightening, since that narrowness comes from the current step ordering rather than from the design — reorder later and it widens quietly. Cheapest fix is probably moving steps 16–17 above this gate: nothing in commit needs the VM, and step 9 has already reset. Alternatively execute_test already stops the VM in a finally: block, and the reset could live there too — that's where cleanup which must always run belongs.

docs/features/tcb-config-commit.md
```diff
@@ -0,0 +1,119 @@
...
+
+We can only test the **no-op commit**. On a normal (non-provisional) host `CommittedTcb == CurrentTcb`, so `snphost commit` has nothing to commit: it leaves the floor where it is and — despite ABI §8.3 — does *not* reset `ReportedTcb`, so any `SNP_CONFIG` override remains in effect (verified empirically against snphost 0.7.0). The `ReportedTcb := CurrentTcb` reset described in the ABI is only observable when commit actually commits a provisionally-loaded firmware image, which requires kernel `DOWNLOAD_FIRMWARE_EX` support that is not yet generally available. The test therefore asserts only that a no-op `commit` succeeds; it does not assert the override-clearing side effect. Teardown stops the VM and runs a final `config reset` regardless of outcome.
```
markg-github, line 97: Related to https://github.com/AMDEPYC/sev-certify/pull/283/changes#r3830772402

"regardless of outcome" isn't quite right. Both teardown steps are type="info", but they sit after the type="setup" commit gate, and a setup failure skips all remaining steps — so neither runs when that gate fails.

The VM half stays true, but for a different reason than this implies: execute_test stops it in a finally: ... stop_vm(launch), not via the Stop VM step. The config reset has no equivalent backstop.

Same issue two sentences up, on line 95: "halting always leaves TCB state clean" holds only if step 9's config-reset succeeded. If that's what failed, halting at the gate leaves the override in place.

Suggest softening both sentences, or moving the teardown ahead of the gate so they become true.

sev_verify/cert_tests/c3_0/c3_0_0_1/snphost_config_commit.py
```diff
@@ -55,15 +80,47 @@ def _run_snphost_tcb() -> subprocess.CompletedProcess:
...
+        raise RuntimeError(
+            f"snpguest display report failed: {proc.stderr.strip()}"
+        )
+    sections = _parse_tcb_sections(proc.stdout)
+    for name in ("Current", "Committed", "Reported"):
+        if name not in sections:
+            raise RuntimeError(
+                f"no {name} TCB section in snpguest display report output"
+            )
+    return {name: sections[name] for name in ("Current", "Committed", "Reported")}
```
DGonzalezVillal, line 120: In Mark's derived key PR he also adds a lot of code for TCB parsing and comparing and verifying. I don't want to hold off PRs because of this, but we should start finding pieces that are repeated through out tests and create helper files we can use in different tests.

sev_verify/cert_tests/c3_0/c3_0_0_1/snphost_config_commit.py
```diff
@@ -106,6 +163,206 @@ def verify_differ(_ctx: StepContext) -> StepHandlerResult:
...
+    return _verify_guest_tcb(str(report_path), expect_match_reported=False)
+
+
+def verify_guest_report_restored(ctx: StepContext) -> StepHandlerResult:
+    """Verify the live-VM report carries the restored (original) TCB values.
+    """
+    report_path = ctx.artifact_dir / "report_after_reset.bin"
+    if not report_path.exists():
+        return StepHandlerResult(exit_code=1, stderr=f"report not found: {report_path}")
+    return _verify_guest_tcb(str(report_path), expect_match_reported=True)
```
DGonzalezVillal, line 238: Could this be one function, you just pass in the name of the report you want to verify? Can we not pass in parameters into the callable functions?

sev_verify/cert_tests/c3_0/c3_0_0_1/snphost_config_commit.py
```diff
@@ -106,6 +163,206 @@ def verify_differ(_ctx: StepContext) -> StepHandlerResult:
     return _verify_result("verify-differ")
```
DGonzalezVillal, line 163: Similar thing here. Does our callable steps not allow passing parameters to the function? This 3 functions could be 1 with a verify parameter.

DGonzalezVillal (commented):
The test and the PR look good. I just wanted to have a discussion about the disposable-host flag. 

I agree that we need a flag to gate tests that may change the state of the machine, like commit and other tests in the future. I'm just don't know if disposable is the right name? It makes it sound that the test can only be run on servers that will be thrown away. I would rather the flag be more of a warning, like dangerous-tests or host-modifying, that let's people know they can be run, but under their own risk. 

I also think we should track somewhere what tests get enabled when the flag is passed. Just so that we know what type of changes could happen if the flag is used. Maybe do that per-certification level. Also people know what to expect if they do pass the flag. 

There's also some overlap between you and Mark's PRs on handling the TCB. I think this will probably not be the last time we have to handle and compare TCBs, so maybe creating a module for shared functions everyone can use could be helpful. But that's out-of-scope for this PR. 

Lastly, does our current design not allow us to pass parameters into our callable functions? If that's the case that's another improvement, we need to work on. I realized you cleverly created work arounds for this, but it would be nice that we don't have to create several functions that do the same thing, just named slightly differently. But that can also be oos for this PR.

For the time being, I want to focus on the new flag.

modules/test/host/sev-verify/mkosi.extra/usr/local/lib/scripts/run-sev-verify.sh
amd-aliem: Yeah that's true, what do you about `--allow-host-changes` to be more specific?

amd-aliem:
Thanks @DGonzalezVillal and @markg-github for the reviews! I've pushed a version that addresses the comments, and opened these 2 issues for follow-up work:
[Issue#304 Create Shared TCB Module](https://github.com/AMDEPYC/sev-certify/issues/304)
[Issue#305 Pass parameters to callable steps](https://github.com/AMDEPYC/sev-certify/issues/305)
Ready for your re-review!

DGonzalezVillal (approved):
LGTM Thank you Amanda!

sev_verify/cli.py
```diff
@@ -290,6 +298,36 @@ def _section(label: str) -> None:
...
+            continue  # a malformed manifest surfaces later in the run loop
+        for test in cert.tests:
+            if test.host_changes:
+                by_level.setdefault(test.level or cert.version, []).append(test.name)
+
+    if not by_level:
+        return
+
+    state = "enabled" if allow_host_changes else "disabled (tests skipped/downgraded)"
+    _flush(f"   Host changes: --allow-host-changes {state}")
```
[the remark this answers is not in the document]
amd-aliem: implemented within the step, not test-wide

sev_verify/models.py
```diff
@@ -276,6 +276,11 @@ class TestDefinition:
     scope: Scope
     description: str = ""
     level: str = ""  # certification level, e.g. "3.0.0-0"
+    # True when the test may change host-level state (e.g. firmware TCB
+    # settings). Such changes are gated behind ``--allow-host-changes`` and
+    # are boot-session-only (reset on reboot). Surfaced at startup so operators
+    # know which tests can touch the host. See StepContext.allow_host_changes.
+    host_changes: bool = False
```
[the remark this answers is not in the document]
amd-aliem: Implemented within the step, not test-wide

sev_verify/cert_tests/c3_0/c3_0_0_1/snphost_config_commit.py
```diff
@@ -106,6 +163,206 @@ def verify_differ(_ctx: StepContext) -> StepHandlerResult:
...
+    )
+
+
+def verify_guest_report_lowered(ctx: StepContext) -> StepHandlerResult:
+    """Verify the fresh-VM report carries the lowered TCB values.
+    """
+    report_path = ctx.artifact_dir / "report.bin"
+    if not report_path.exists():
+        return StepHandlerResult(exit_code=1, stderr=f"report not found: {report_path}")
+    return _verify_guest_tcb(str(report_path), expect_match_reported=False)
```
markg-github: No report.bin staleness detection/incomplete staleness prevention

artifact_dir uses exist_ok=True and is never cleaned. test_artifact_dir returns a deterministic artifacts_root/<version>/<level>/<test_name> with no run ID; no rmtree, shutil, or unlink appears anywhere in cli.py, runner.py, guest_vsock.py or models.py. A failed required step does not break the loop — only setup does — so steps 7 and 8 run. run_guest_pull_step never touches the destination on failure (no pre-truncate, no finally), so a prior run's report.bin survives intact.

Steps 7 and 8 affirmatively PASS on the stale file. Step 2 recomputes the lowered TCB deterministically from the live Platform TCB, so a re-run on unchanged firmware reproduces the exact values the stale report was generated under. Nothing checks report freshness — step 5 uses --random, but request.bin is never read back. Step 8 then does a live KDS fetch and verify attestation --signature, both of which succeed on any validly-signed report.

The same pattern appears at four call sites, not one: lines 226-228, 235-237, 248-249, and 298-300.

Conditions: a prior run left report.bin in the same dir; this run's step 4, 5 or 6 fails; firmware is unchanged between runs. All three are routine on a developer host iterating on a failing VM launch. The sanctioned CI path is insulated — run-sev-verify.sh runs on a fresh image and never passes --artifacts-dir — so the exposure is the manual re-run path. The overall verdict is still fail, so this is misleading step-level evidence rather than a false pass.

sev_verify/cli.py
```diff
@@ -370,6 +409,7 @@ def execute_test(
             launch=None,
             cli_qemu_binary=qemu_binary,
             cli_ovmf_path=ovmf_path,
+            allow_host_changes=allow_host_changes,
```
markg-github: Does the test's host_changes attribute do everything it's supposed to do? It doesn't seem to affect runtime behavior, just what gets printed, ie, there's no enforcement.

sev_verify/cli.py
```diff
@@ -665,6 +715,8 @@ def main(argv: list[str] | None = None) -> int:
         _flush(f"   OVMF:   {effective_ovmf}")
     _flush("")
 
+    _print_host_changes_notice(manifest_entries, args.allow_host_changes)
```
markg-github: Having this here misses any prereq tests (see the lines that follow this one).

sev_verify/cli.py
```diff
@@ -290,6 +298,36 @@ def _section(label: str) -> None:
...
+        except ValueError:
+            continue  # a malformed manifest surfaces later in the run loop
+        for test in cert.tests:
+            if test.host_changes:
+                by_level.setdefault(test.level or cert.version, []).append(test.name)
+
+    if not by_level:
+        return
+
+    state = "enabled" if allow_host_changes else "disabled (tests skipped/downgraded)"
```
markg-github: Agree with the Copilot comment

markg-github (commented):
As promised, remaining, post-merge comments.

PR #297, opened 2026-08-19
Fix: Update Certificate Workflow to create PR for opensuse-leap OS Release

LakshmiSaiHarika:
Earlier, in the Opensuse issues, the GH Action for the certificate workflow was missing, which is fixed in this PR.

In this PR, I added a fix for the regex expression to include the expression 'opensuse-leap' in issues.
Also, I included `milestoned` event to trigger the GH Action to update the certificate when a milestone is updated.

Refer to my updated opensuse issue in my forked repo to show the GH action for the cert workflow update: https://github.com/LakshmiSaiHarika/snpcert/issues/130
GH Bot generated PR Link: https://github.com/LakshmiSaiHarika/snpcert/pull/129

### DIFF
--- a/.github/workflows/update-certification-matrix.yml
+++ b/.github/workflows/update-certification-matrix.yml
@@ -23,7 +23,7 @@ jobs:
             
             // Extract and validate labels
             const procLabel = issue.labels.find(l => l.name.match(/^proc-\d+$/i));
-            const distroLabel = issue.labels.find(l => l.name.match(/^os-[a-zA-Z]+(?:-[a-zA-Z0-9.]+)?$/i));
+            const distroLabel = issue.labels.find(l => l.name.match(/^os-[a-zA-Z]+(?:-[a-zA-Z0-9.]+)*$/i));
             if (!procLabel || !distroLabel) return console.log('Missing processor or OS label');
             
             const procSeries = procLabel.name.replace(/^proc-/i, '');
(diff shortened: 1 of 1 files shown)


.github/workflows/update-certification-matrix.yml
```diff
@@ -2,7 +2,7 @@ name: Update Certification Matrix
 
 on:
   issues:
-    types: [opened]
+    types: [opened, milestoned]
```
amd-aliem: Uncertain if we need this 'milestoned' trigger since dispatch attaches the milestone as a part of opening the issue? It might generate duplicate workers (if people add and remove milestones from a certificate, which they shouldn't do). Was this fixing a bug you were seeing?

LakshmiSaiHarika: If we don't want duplicate workers, I am ok with it. I refactored my code and re-tested with the opensuse to check the GH bot actions for this OS release. 

I will attach the opensuse cert link of my forked repo, along with GH bot PR for opensuse.

I just added this `milestoned` event because  I thought users could update the milestone of the same certificate if required, and to reflect this change in the cert table.

LakshmiSaiHarika: Please look at my forked issue for the latest OpenSUSE certificate with GH bot PR step( creates PR to add opensuse in the cert table): https://github.com/LakshmiSaiHarika/snpcert/issues/131

amd-aliem: ok sounds good!

